PR #82, opened 2025-11-17
12 workspace setup

BenBotros:
## What does this change?

Adds an onboarding page that collects the user's unit and year preferences before allowing access to the application. The page adapts dynamically based on available options:

- If only 1 unit exists: shows year selection only
- If only 1 year exists: shows unit selection only
- If only 1 unit and 1 year exist: automatically redirects to home
- Selections are persisted in cookies for subsequent logins

## Why is this needed?

Users need to specify their context (unit/year) before accessing the application. This ensures the app can display relevant data from the start rather than requiring users to configure this mid-workflow. By storing preferences in cookies, returning users skip this step entirely, improving the experience for repeat visits.

## Type of change

- [x] 🐛 Bug fix
- [x] 🔧 Configuration change

## How to test this

## Steps to test:
- Clear cookies and log in → should land on onboarding page
- Select unit and year → should redirect to home and store cookies
- Log out and log back in → should bypass onboarding and go directly to home
- Test with mock data having: only 1 unit, only 1 year, or only 1 unit+year combination
- After onboarding is complete, verify you can still navigate back to change unit/year selections
- Attempt to access protected routes without completing onboarding → should be blocked by guard

### DIFF
--- a/frontend/src/pages/app/WorkspaceSetupPage.vue
+++ b/frontend/src/pages/app/WorkspaceSetupPage.vue
@@ -1,5 +1,5 @@
 <script setup lang="ts">
-import { ref, computed, watch, onMounted } from 'vue';
+import { ref, computed, onMounted } from 'vue';
 import { useWorkspaceStore } from 'src/stores/workspace';
 import { useRouter, useRoute } from 'vue-router';
 import LabSelectorItem from 'src/components/organisms/workspace-selector/LabSelectorItem.vue';
@@ -23,10 +23,6 @@ const selectedUnitAffiliations = computed(() => {
 
 const hasMultipleYears = computed(() => {
   return (workspaceStore.unitResults?.years.length ?? 0) > 1;
-});
-
-const hasSingleYear = computed(() => {
-  return workspaceStore.unitResults?.years.length === 1;
 });
 
 const yearsCount = computed(() => {
--- a/frontend/src/stores/workspace.ts
+++ b/frontend/src/stores/workspace.ts
@@ -123,7 +123,7 @@ export const useWorkspaceStore = defineStore(
         unitResultsErrors.value = [];
 
         unitResults.value = (await api
-          .get(`unit/${id}/results`)
+          .get(`unit/${id}/results`, { searchParams: options })
           .json()) as UnitResults;
       } catch (error) {
         console.error('Error getting unit results:', error);
--- a/Makefile
+++ b/Makefile
@@ -98,13 +98,11 @@ ci: ## Run all CI checks (use before pushing to dev/stage/main)
 
 .PHONY: format
 format: ## Format all code (root, backend, frontend, docs, helm)
-	@echo "Formatting root files..."
+	@echo "\033[36mFormatting root files...\033[0m"
 	@npx prettier --write "*.md" "*.json" "*.yml" --ignore-unknown --ignore-path .prettierignore
-	@if [ -d "backend" ]; then cd backend && $(MAKE) format FILES="."; fi
-	@if [ -d "frontend" ]; then cd frontend && $(MAKE) format FILES="."; fi
-# 	@if [ -d "docs" ]; then cd docs && $(MAKE) format FILES="."; fi
-# 	@if [ -d "helm" ]; then cd helm && $(MAKE) format FILES="."; fi
-	@echo "✅ Formatting complete!"
+	@if [ -d "backend" ]; then $(MAKE) -C backend format FILES="."; fi
+	@if [ -d "frontend" ]; then $(MAKE) -C frontend format FILES="."; fi
+	@echo "\033[32m✅ Formatting complete!\033[0m"
 
 # =============================================================================
 # Linting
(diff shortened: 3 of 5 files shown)


frontend/src/stores/workspace.ts
```diff
@@ -0,0 +1,177 @@
...
+        units.value = allUnits
+          .map((unit) => ({
+            ...unit,
+            role: user.roles.find(
+              (r) =>
+                typeof r.on === 'object' &&
+                String(r.on.unit) === String(unit.id),
+            )?.role,
+          }))
+          .filter((u) => u.role);
```
BenBotros: @guilbep  Need to discuss if this mapping is necessary or if it can be done in backend. According to our previous discussion it would complicate the design of the API
Screenshot 2025-11-17 at 18 34 05

frontend/src/stores/workspace.ts
```diff
@@ -0,0 +1,190 @@
+import { defineStore } from 'pinia';
+import type { PersistenceOptions } from 'pinia-plugin-persistedstate';
+import { ref, computed } from 'vue';
```
guilbep: @BenBotros 


- [ ] Use centralized API endpoints and remove hardcoded URLs.
    - [ ] Use `api` import for API calls.
- [ ] fetch is get so replace fetchUnits by getUnits, let's use HTTP VERB get/post/path/delete
- [ ] Rewrite `fetchUnit` to fetch a single unit by ID.
- [ ] Use the `options` parameter in `fetchUnitResults` for pagination/sorting.
- [ ] try catch with empty catch is bad practice errors in `initFromPersisted` for better debugging.
    -> let's. think of a way of handling errors globally maybe ? 


- [ ] Remove console logs or replace with proper logging.
- [ ] Clean up redundant or obvious comments.
- [ ] Use separate loading states for different fetch actions.
     -> const units
     -> const unitsLoading
     -> const unitsErrors []

@guilbep maybe later in new issue https://github.com/EPFL-ENAC/co2-calculator/issues/118
- [ ] Ensure interfaces match backend responses and are documented.
    --> maybe later @guilbep (https://openapi-ts.dev/openapi-fetch/ + openapi-typescript)
- [ ] Add retry or fallback logic for failed fetches.
  -> todo later for @guilbep

frontend/src/pages/app/WorkspaceSetupPage.vue
```diff
@@ -1,18 +1,235 @@
...
+        </q-banner>
+      </div>
+      <div v-else class="q-mt-sm column-grid-2">
+        <LabSelectorItem
+          v-for="unit in unitsWithRoles"
+          :key="unit.id"
+          :selected="selectedLab === unit.id"
+          :unit="unit"
+          @click="
+            async () => {
```
guilbep: @BenBotros needs to be in the <script> above and named!

frontend/src/pages/app/WorkspaceSetupPage.vue
```diff
@@ -1,18 +1,235 @@
...
+              selectedLab = unit.id;
+              workspaceStore.setUnit(unit);
```
guilbep: @BenBotros not sure I understand what you do here. But I'd like to avoid this behavior as much as possible.

URL state should reflect what we do

frontend/src/pages/app/WorkspaceSetupPage.vue
```diff
@@ -1,18 +1,235 @@
...
+            }
+          "
+        />
+      </div>
+    </q-card>
+
+    <!-- Year Selection -->
+    <q-card
+      v-if="
+        selectedLab !== null &&
```
guilbep: @BenBotros  it's getting complicated, same as above, we can use a computed for that

v-if="hasUnitResulsts"

frontend/src/pages/app/WorkspaceSetupPage.vue
```diff
@@ -1,18 +1,235 @@
...
+          workspaceStore.unitResults.years.map((y) => ({
+            year: y.year,
+            progress: 100,
+            completed_modules: y.completed_modules,
+            comparison: y.last_year_comparison ?? null,
+            kgco2: y.kgco2,
+            status: 'complete',
+          }))
+        "
+        @select="selectedYear = $event"
```
guilbep: @BenBotros would be nice if YearSelector used v-model

frontend/src/pages/app/WorkspaceSetupPage.vue
```diff
@@ -1,18 +1,235 @@
...
+      <div class="column-grid-2">
+        <q-card flat v-if="unitsWithRoles.length > 1" class="container">
+          <h6 class="text-h6 text-weight-medium">
+            {{ $t('workspace_setup_confirm_lab') }}
+          </h6>
+          <h3 class="text-h3 text-weight-medium q-pt-sm">
+            {{ workspaceStore.selectedUnit?.name }}
+          </h3>
+          <p class="text-caption">
+            {{ workspaceStore.selectedUnit?.affiliations.join(' / ') }}
```
guilbep: @BenBotros  computed instead

frontend/src/pages/app/WorkspaceSetupPage.vue
```diff
@@ -1,18 +1,235 @@
...
+        <q-btn
+          color="accent"
+          :label="$t('workspace_setup_confirm_selection')"
+          unelevated
+          no-caps
+          size="md"
+          class="text-weight-medium"
+          @click="
+            workspaceStore.setYear(selectedYear!);
+            router.push({
```
guilbep: @BenBotros  use, :to instead of @click 

and why do we setup store ?

frontend/src/components/layout/Co2Header.vue
```diff
@@ -30,6 +30,32 @@ const handleLogout = async () => {
       <q-space />
 
       <Co2LanguageSelector />
+
+      <template v-if="route.name !== 'workspace-setup'">
+        <span
+          v-if="route.params.unit && route.params.year"
+          class="text-body2 text-weight-medium q-ml-xl q-mr-sm"
+        >
+          {{ decodeURIComponent(route.params.unit as string) }} |
```
guilbep: @BenBotros computed!

frontend/src/css/05-components/_organisms/_table.scss
```diff
@@ -0,0 +1,42 @@
+@use '../../02-tokens' as tokens;
+
+.co2-table {
+  border: 1px solid tokens.$table-border-color;
```
guilbep: @BenBotros  let's double check that all of this is necessary again

frontend/src/stores/store-flag.d.ts
```diff
@@ -1,9 +0,0 @@
-// THIS FEATURE-FLAG FILE IS AUTOGENERATED,
-//  REMOVAL OR CHANGES WILL CAUSE RELATED TYPES TO STOP WORKING
```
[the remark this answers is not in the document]
BenBotros: @guilbep I don't have this file

Issue #51, opened 2025-11-06
[FEAT](Deployment) - DSI infrastructure


charlottegiseleweil:
## 🎯 Feature Overview 
Deployment of dev & stage environments - on DSI infrastructure - see [Contributing.MD](https://github.com/EPFL-ENAC/co2-calculator/blob/main/CONTRIBUTING.md)

## 🚨 Blocking points
- [ ] Security strategy -- strategy to mitigate risk of hostile tenant, since multi-tenancy system ?

### Ongoing  

- [x] Understanding of DSI infrastrucuture
- [x] Access to DSI infra.
- [x] Deploiement DEV  ([osd](go.epfl.ch/osd) ) : @nicdub to deploy app in dev, based on the k8s config in enac repo (helm chart). Make a specific git => https://github.com/EPFL-ENAC/openshift-app-config => https://co2-calculator-dev2.epfl.ch/en/login
- [x] Deploiement STAGE (view resources avail . via [ost](go.epfl.ch/ost) ) => https://co2-calculator-stage2.epfl.ch/en/login
- [x] Implemented DB dumps to PVC via cronjob
- [x] Deploiement (pre)PROD (view resources avail . via [osp](go.epfl.ch/osp) ) => à demander @nicdub 
- [x] SLA / Document expected downtimes (e.g. event du 12.11). 
- [x] Infrastructure resources validated (CPU/mem/network thresholds)

- [x] Cluster Setup | OpenShift cluster provisioned and accessible with proper RBAC (role-based access control) | Dev / Stage => Compte de service géré par DSI. À documenter. 

### Postponed. / In progress
- [x] Link to the right DBs in test/prod environments. 
- [x]  Networking : Routes, services, and ingress configurations are functional; internal/external DNS resolves correctly. OpenShift routes deployment included in Helm chart
- [x] CI/CD Integration : Pipelines (Jenkins, Tekton, or GitHub Actions) automatically trigger build → test → deploy on commits #113 
- [x] Monitoring => #116 .
- [x] Alterting 
- [ ] Logging (INTEGRATION Grafana -> Transfert des logs applicatifs sur infra persistante) ->Move to S5 / Interface Service mgr -- @nicdub en cours. 
- [ ] Extraction des logs d'observabilité - selon discussion #240 
- [ ] Rollback tested and documented - QVG ?
[...]

PR #147, opened 2025-11-26
chore(backend): add logger for health endpoint

guilbep:
## What does this change?
This pull request adds improved logging to the `health` endpoint in `backend/app/main.py`. Now, every health check request is logged with relevant details, and the log level reflects the health status.

Observability improvements:

* Added structured logging to the `health` endpoint, including health status, database status, accreditation status, and additional details. The log level is set to `info` for healthy checks and `warning` for unhealthy checks.

- [x] 🐛 Bug fix

## How to test this

- [ ] Steps to test: call localhost:8000/health and look at the logs

## Related issues

- Related to #51

### DIFF
--- a/backend/app/main.py
+++ b/backend/app/main.py
@@ -144,32 +144,47 @@ async def health():
         db_status = "error"
         details["db_error"] = str(e)
 
-    # Accred (role_provider) check
-    try:
-        import httpx
+    # Role provider health check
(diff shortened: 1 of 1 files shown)


ymarcon (approved):
Very EPFL specific with this accred reference, but I guess the problem is wider than the scope of this PR.

Issue #213, opened 2025-12-04
[FEAT](Results) Front-end full layout including Big Numbers


charlottegiseleweil:
### 🏷️ Title Section

<img width="1283" height="127" alt="Image" src="https://github.com/user-attachments/assets/507602f0-c5d5-43c6-9e9a-58ce0fb83b27" />

_(Design is not totally correct, it gives a draft idea)_

- [x] Title: "Results"/"Résultats" (EN/FR)
- [x] Subtitle: "Annual carbon footprint {YYYY}" / 
- [x] `Download as PDF` button downloads this page as PDF (printing this html is ok for now)
- [x] Checkbox `View Uncertainties` → Uncertainty tags and error bars appear. 

### Big numbers section

<img width="1201" height="150" alt="Image" src="https://github.com/user-attachments/assets/094c148a-2f1d-44e0-bf97-67f7893ffad7" />

_(Design is not totally correct, it gives a draft idea)_

- [x] Three big numbers displayed in one line, each in a box. 
- [x] `Total Unit Carbon Footprint`:  
  - Number: the total amount of `t CO₂-eq` → The sum of the total of all modules + the Additional Categories (processed with the total FTE of the lab)  
  - Side-Text: is equivalent to {km} driven by car, calculated with `0.34 kg CO₂-eq / km`  
  - Info button: "Calculated with the value of 0.34 kg CO₂-eq / km"
- [x] `Carbon Footprint per FTE`:  
  - Number: the total amount of `t CO₂-eq` divided by the total FTE of the unit  
  - Text: The total carbon budget per person according to the Paris Agreement is 2 t CO₂-eq  
  - Info button: "Following the Paris Agreement."
- [x] `Unit Carbon Footprint compared to <YYYY>`: nicetohave : possibility to change the year. default is the previous year. 
  - Number: % of the total difference with the selected year  
  - Text: Compared to {YYYY} value of {total_value_selected_year} | if no comparison year available → "No comparison year available"

### Result by category Title section

<img width="1201" height="76" alt="Image" src="https://github.com/user-attachments/assets/aea09e01-788c-4047-bb74-a6a864506479" />

- [x] Title: Results by Category
- [x] Subtitle: Annual Carbon Footprint 2024
- [x] All Modules are expanded by default but can be collapsed

## 💾 Sample data
[...]

PR #293, opened 2025-12-17
fix/213 Add uncertainty visualization toggle and update result page labels

BenBotros:
## What does this change?
- Adds optional uncertainty visualization to all carbon footprint charts with a toggle control
- Implements uncertainty bands for each category (displayed with ±values in tooltips)
- Updates result page labels to be module-specific instead of generic "Unit" labels
- Makes year dynamic throughout the app by pulling from workspace store instead of hardcoded "2024"
- Simplifies equipment name label (removes redundant module title prefix)
- Fixes comparison values (0.34t instead of 0.34 kg)
- Adds helper function to convert hex colors to rgba for uncertainty opacity styling

## Why is this needed?
- Users need to understand the confidence levels and uncertainty ranges in carbon footprint calculations
- Dynamic year selection improves flexibility for multi-year assessments
- Module-specific labels provide clearer context when viewing individual module results
- Consistent units (tonnes vs kg) improve data accuracy and user understanding

## Type of change

- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [] 📝 Documentation update
- [ ] 🎨 Design/UI improvement
- [ ] 🔧 Configuration change
- [ ] 🧹 Code cleanup

## Testing checklist

- Related to #213

### DIFF
--- a/frontend/src/components/layout/Co2Header.vue
+++ b/frontend/src/components/layout/Co2Header.vue
@@ -73,7 +73,9 @@ const isInBackOfficeRoute = computed(() => isBackOfficeRoute(route));
     <q-toolbar class="q-px-xl q-py-md">
       <q-toolbar-title class="row items-center no-wrap">
         <router-link
-          :to="{ name: 'home', params: route.params }"
+          :to="{
+            name: 'root',
+          }"
           class="toolbar-home-link row items-center no-wrap"
         >
           <q-img src="/epfl-logo.svg" :alt="$t('logo_alt')" width="100px" />
--- a/frontend/src/components/organisms/workspace-selector/LabSelectorItem.vue
+++ b/frontend/src/components/organisms/workspace-selector/LabSelectorItem.vue
@@ -7,7 +7,7 @@ import { useWorkspaceStore } from 'src/stores/workspace';
 const props = defineProps<{
   selected?: boolean;
   unit: {
-    id: number;
+    id: string;
     name: string;
     principal_user_id: string;
     principal_user_function: string;
--- a/frontend/src/pages/app/HomePage.vue
+++ b/frontend/src/pages/app/HomePage.vue
@@ -20,21 +20,16 @@ const modulesCounterText = computed(() =>
 );
 
 const homeIntroWithLinks = computed(() => {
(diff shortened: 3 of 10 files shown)


frontend/src/constant/charts.ts
```diff
@@ -123,17 +123,31 @@ export const chartColorScale: ColorColumn[] = [
   },
 ];
 
+// Helper function to convert hex color to rgba with opacity
+const hexToRgba = (hex: string, opacity: number): string => {
+  const result = /^#?([a-f\d]{2})([a-f\d]{2})([a-f\d]{2})$/i.exec(hex);
+  if (!result) return hex;
+  const r = parseInt(result[1], 16);
```
guilbep: @BenBotros  I think we can do better than this

https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/hex-color

hex colors are supporting 8 char, you could add opacity just by converting [0,1] to 255 hex value add concat the string of your colors.

bottom line, that's too complex for just adding opacity + it seems that you either put 0.5 or 0 in opacity..
you don't even need to compute anything.

I'd remove this function call altogether and maybe store the opacity 0.5 version already in the array you created.

frontend/src/constant/charts.ts
```diff
@@ -123,17 +123,31 @@ export const chartColorScale: ColorColumn[] = [
...
+  const g = parseInt(result[2], 16);
+  const b = parseInt(result[3], 16);
+  return `rgba(${r}, ${g}, ${b}, ${opacity})`;
+};
+
 export const getElement = (
   name: string,
   shade: number = 2,
```
guilbep: @BenBotros it's weird that the shade is a number

In the future let's be explicit with maybe key, value, it's too brittle: getElement('', 0, 0.5) is weird
we could have something like

Rename the function getColorShade({
 colorName: 'lightGrey',
 shade: 'medium',
 opacity?: 0.5,
 colorBlind: false
})
```
colors: {
'dark': '#000',
 'medium' | 'light' | 'lighter' | 'pale' .. etc 
}
```

And also stop using chartColorScale.find!!!

PR #317, opened 2026-01-05
fix/student-counter-corrections

BenBotros:
## What does this change?
This PR transforms the Student FTE Calculator from a manual submission workflow to an automatic calculation system. The calculator now instantly updates the parent form as users input values, removing the need for explicit "submit" actions. Additionally, the student time field in the my-lab module now supports sorting and inline editing.

## Why is this needed?
The previous workflow required users to manually click "Use This Value" after calculating FTE, creating unnecessary friction. By automatically updating values as they're calculated, we streamline the user experience and reduce the number of steps needed to complete the form. The inline editing and sorting capabilities further improve usability by allowing users to modify values directly in the table.

- Related to #294

### DIFF
--- a/frontend/src/components/organisms/module/StudentFTECalculator.vue
+++ b/frontend/src/components/organisms/module/StudentFTECalculator.vue
@@ -1,10 +1,6 @@
 <template>
   <div>
-    <q-form
-      ref="formRef"
-      class="student-fte-calculator q-mx-md"
-      @reset="onReset"
-    >
+    <q-form ref="formRef" class="student-fte-calculator q-mx-md">
       <q-input
         ref="studentsRef"
         v-model.number="students"
@@ -43,7 +39,7 @@
 </template>
 
 <script setup lang="ts">
-import { ref, computed, nextTick, watch } from 'vue';
+import { ref, computed, watch } from 'vue';
 import { useI18n } from 'vue-i18n';
 import type { QForm } from 'quasar';
 
@@ -52,7 +48,6 @@ const { t: $t } = useI18n();
 const students = ref<number | null>(null);
 const duration = ref<number | null>(null);
 const avgFTE = ref<number | null>(null);
-
 const formRef = ref<QForm | null>(null);
 
 const emit = defineEmits<{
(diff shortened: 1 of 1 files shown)


frontend/src/components/organisms/module/StudentFTECalculator.vue
```diff
@@ -39,33 +38,12 @@
         outlined
         dense
       />
-
-      <div class="calculated-result flex justify-between items-center">
-        <div>
-          <span class="text-body2 text-secondary"
-            >{{ $t('student_helper_calculated_label') }}:</span
```
guilbep: @BenBotros What happen to the text here btw ?

not sure @reset is ever called here ? please remove if not used

frontend/src/components/organisms/module/StudentFTECalculator.vue
```diff
@@ -88,11 +66,16 @@ const calculatedFTE = computed(() => {
...
-    return '-';
+// Automatically update the parent form when calculation is valid and all fields are filled
+watch(calculatedFTE, (newValue) => {
+  if (
+    newValue > 0 &&
+    students.value !== null &&
+    duration.value !== null &&
+    avgFTE.value !== null
+  ) {
+    emit('use-value', Number(newValue.toFixed(1)));
```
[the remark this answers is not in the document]
guilbep: ``
const isCompleteAndValid = computed(() => {
  return (
    students.value !== null &&
    duration.value !== null &&
    avgFTE.value !== null &&
    students.value > 0 &&
    duration.value > 0 &&
    duration.value <= 12 &&
    avgFTE.value > 0 &&
    avgFTE.value <= 1
  );
});
Step 2 — watch transition to valid + value change
let lastEmitted: number | null = null;

watch(
  [isCompleteAndValid, calculatedFTE],
  ([valid, fte]) => {
    if (!valid) return;

    const rounded = Number(fte.toFixed(1));

    if (rounded !== lastEmitted) {
      lastEmitted = rounded;
      emit('use-value', rounded);
    }
  },
  { flush: 'post' } // wait for DOM + model to settle
);
Why this works
✔ No debounce
✔ No submit
✔ No spam
✔ Emits only meaningful changes
✔ Deterministic (no timers)
This is the correct reactive equivalent of submit.

Important: why flush: 'post' matters
Without it:
watcher may run mid-input
validations may still be settling
With it:
Vue batches updates
watcher runs once per tick
keystroke spam collapses naturally
This is a Vue-native stabilization mechanism (and often overlooked).
Optional UX refinement (highly recommended)
Reset the emission when the form resets:
function onReset() {
  students.value = null;
  duration.value = null;
  avgFTE.value = null;
  lastEmitted = null;

  nextTick(() => {
    formRef.value?.resetValidation();
  });
}

```

@BenBotros  NO debounce! and inspire yourself from the avobe

frontend/src/components/organisms/module/StudentFTECalculator.vue
BenBotros: we decided to remove it with @charlottegiseleweil as it was redundant with the input below

it isn't used i removed it

guilbep (approved):
Not tested,

PR #532, opened 2026-03-02
feat/483 introduce separate tables for plane and train locations

### LINKED ISSUE #483: [FEAT](Travel) Refactor as submodules
# 📌 Problem

The current `Travel` module data is incoherent compared to other modules, because it mixes Plane and Train data that are in different formats. 

- The datasets are not in the same formats, do not use the same emission factors. Back Office treats them differently (2 rows)

- But The module UI treats them as a single table

➡ Result: Inconsistent architecture.
This creates confusion / incoherence with other modules. 

# 💡 Proposed Solutions

## ✅ Option 1 – Split into 2 Submodules (Recommended)

Separate the Travel module into 2 submodes (comme `IT Equipments` and `Scientific Equipments`) :  `Plane Trips` & `Train Trips` - which can each have their own specific columns and their own emission factors. 

- Back Office / Data Management : 2 lines for trains and planes, as requested per @martina-gallato 
- Travel Module : **2 distinct tables**, with their own "Upload CSV" by the user. 

**+ Benefits:** Data formats of trains and plane can differ.  Cleaner data structure & Better scalability

( Recommended by IT4R and @martina-gallato )

## ⚖️ Option 2 – Keep a Single Submodule (Requires Data Harmonization)

Merge Plane and Train datasets in a single submodule, making all columns consistent (Unified schema for planes and trains)

- Back Office / Data Management : A single line to upload both planes and trains. Must merge emission factors
- Travel Module : Samesame. Upload CSV button can mix planes and trains - we add the column `"transport_mode"`

## 🤯 Option 3 : Current Situation (inconsistent)

There is a structural inconsistency between Back Office and Module behavior:

- Back Office / Data Management : Trains and planes Treated as 2 submodules (@martina-gallato requires 2 separate lines to upload data and emission factors)
- Travel Module : Treated as 1 single submodule, since you wanted all trips in one table. -> It's a problem for users uploading CSVs because it's not the same CSV upload for planes and trains !

-> Fyi, it's not impossible to keep it as is. It's always possible to "hack" if you really want to keep it as is, according to your current specs, but it will be less consistent/elegant for future developers/institutions to pick-up this project and understand this choice. 

# ❓ Clarifications needed 
Pinging you @agletiec (en tant que propio de la spec travel) et @CDucrest-EPFL : on aurait besoin de votre optinion sur ce sujet, et de choisir une solution (si on change pour l'Option 1 ou 2, plus élégantes/cohérentes) mais qui impliquent un changement de specs. 
Désolée pour le franglais.

BenBotros:
## What does this change?
Splits the single locations table into two separate tables — locations_plane (PlaneLocation) and locations_train (TrainLocation) — removing the transport_mode discriminator column entirely. All queries, joins, aliased references, and seed logic that previously filtered on Location.transport_mode now select from the appropriate concrete model directly. The location search and distance API endpoints move from a shared /locations/search?transport_mode= query-parameter pattern to dedicated /locations/search/plane, /locations/search/train, and /locations/calculate-distance/{mode} routes, with the frontend updated to match. transport_mode is also removed as a stored field on travel data entries: the mode is now inferred from data_entry_type at query time, so compute_trips and compute_travel derive it from DataEntryTypeEnum rather than reading it from the JSON payload. The seed script is updated to read from two separate CSV files (seed_travel_location_plane.csv and seed_travel_location_train.csv) rather than one combined file with a mode

## Why is this needed?
Storing transport_mode as a column on a shared locations table meant every query had to carry a filter predicate to avoid cross-contaminating plane and train results, and every join had to be verified to include that filter. The discriminator was redundant — a location's mode is fixed at insert time and is always known from context — so it added noise without providing safety. Splitting into two tables makes the constraint structural: it is impossible to accidentally query train stations when looking up airports, the DB engine can use narrower, mode-specific indexes, and the application code no longer needs to carry transport_mode through layers that do not otherwise need it. Removing transport_mode from the travel data entry payload has the same motivation: the entry's data_entry_type already encodes whether the trip is a flight or a train journey, so storing it again in the JSON blob was redundant and a potential source of inconsistency.

## Type of change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📝 Documentation update
- [ ] 🎨 Design/UI improvement
- [ ] 🔧 Configuration change
- [x] 🧹 Code cleanup

## Related issues

- Closes #483 

---


guilbep:
@charlottegiseleweil  @BenBotros  If you're okay. I'm gonna close this now that we're still using the same table for both train and plane

guilbep:
[charlottegiseleweil](https://github.com/charlottegiseleweil)
closed this as [completed](https://github.com/EPFL-ENAC/co2-calculator/issues?q=is%3Aissue%20state%3Aclosed%20archived%3Afalse%20reason%3Acompleted)[on Mar 9](https://github.com/EPFL-ENAC/co2-calculator/issues/483#event-23402844694)

PR #548, opened 2026-03-03
fix/chart corrections

BenBotros:
## What does this change?
Refactors ModuleCarbonFootprintChart to use backend-driven subcategory keys directly (co2, ch4, n2o, refrigerants, clouds, ai, full Purchases breakdown) instead of client-side normalization aliases. Replaces getPrimaryOrSum fallback logic with a simpler normalizeBreakdownEntry that strips metadata/object payloads. Adds CHART_SUBCATEGORY_COLOR_SCHEMES and getChartSubcategoryColor to charts.ts for shared subcategory color resolution. Unvalidated categories are now zeroed out in the chart via isCategoryValidated/zeroNumericValues instead of being filtered out. The emission treemap is rebuilt from a new module_treemap backend field instead of being derived from module_breakdown rows. getEmissionBreakdown gains an includeUnvalidated flag and deduplication logic that accounts for it. The refresh-request mechanism is upgraded from a boolean flag to a monotonic sequence counter to prevent double-consumption.

## Why is this needed?
The previous client-side key normalization (getPrimaryOrSum with fallback arrays) was brittle — it diverged from actual backend key names as the schema evolved, causing silent zero values in charts. Centralizing subcategory color schemes removes duplicated color definitions across components. The sequence-based refresh mechanism fixes a race condition where concurrent watchers could both consume the same refresh request or skip a legitimate one.

## Type of change

- [x] 🐛 Bug fix
- [x]  ✨ New feature
- [ ] 📝 Documentation update
- [ ] 🎨 Design/UI improvement
- [ ] 🔧 Configuration change
- [x] 🧹 Code cleanup
- [ ] 
---

### DIFF
--- a/backend/app/modules/professional_travel/schemas.py
+++ b/backend/app/modules/professional_travel/schemas.py
@@ -17,7 +17,10 @@ from app.schemas.data_entry import (
     DataEntryResponseGen,
     DataEntryUpdate,
 )
-from app.utils.distance_geography import compute_travel_distance_km
+from app.utils.distance_geography import (
+    calculate_plane_distance,
+    calculate_train_distance,
+)
 
 logger = get_logger(__name__)
 
@@ -162,6 +165,7 @@ class ProfessionalTravelPlaneModuleHandler(ProfessionalTravelBaseModuleHandler):
         from origin/destination airports."""
         origin_id = data_entry.data.get("origin_location_id")
         dest_id = data_entry.data.get("destination_location_id")
+        number_of_trips = data_entry.data.get("number_of_trips", 1)
         if not origin_id or not dest_id:
             return {}
         from app.services.location_service import LocationService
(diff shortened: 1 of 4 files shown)


frontend/src/composables/useEmissionTreemap.ts
```diff
@@ -29,31 +61,61 @@ function buildTreemapFromRows(
...
+        value: childNodes.reduce((sum, node) => sum + node.value, 0),
+        children: childNodes,
+      });
+    }
+
+    const parentKeys = new Set(parentToChildren.keys());
+    for (const leaf of leaves) {
+      // Keep flat nodes only when not already represented as parent or child.
+      if (!consumedChildren.has(leaf.name) && !parentKeys.has(leaf.name)) {
+        groupedChildren.push({ ...leaf });
```
guilbep: @BenBotros 

c'est dommage que ce soit fait en frontend, çå devrait être le backend qui te construit la data...

frontend/src/components/charts/GenericEmissionTreeMapChart.vue
```diff
@@ -35,31 +42,191 @@ const props = defineProps<{
...
+  additional: 'charts-other-purchases-subcategory',
+  // External cloud & AI leaves
+  stockage: 'charts-stockage-subcategory',
+  virtualisation: 'charts-virtualisation-subcategory',
+  calcul: 'charts-calcul-subcategory',
+  ai_provider: 'charts-ai-provider-subcategory',
+};
+
+function resolveLabel(raw: string): string {
+  const mappedKey = TREEMAP_LABEL_KEY_MAP[raw];
```
guilbep: 0_0'

PR #867, opened 2026-04-15
fix/850 travel module bug

### LINKED ISSUE #850: [BUG](Travel) Data not displayed in Table
In Stage (release 0.8) : 
I added manuellay a train trip (with "number of trips" = 999) and it appears in calculation, and apparently in DB, but not visibile in front-end table 
Image

Plane tripy also not visible in table: 
Image

BenBotros:
- **fix(travel): repare number of trips**
- **fix: repare poermision errors in travel module**

## What does this change?

This PR fixes core functionality in the Travel module involving data recalculation and access control:

* **Emissions Recalculation:** Updated `DataEntryEmissionService` to strip existing `kg_co2eq` values during updates. This ensures that when a user changes fields like `number_of_trips`, the system recomputes the CO2 instead of using a stale value (common in CSV imports).
* **Permission Logic:** Fixed a mismatch in unit authorization where roles were being compared against `unit.institutional_id` (cost-center) instead of `unit.institutional_code` (Accred provider code).
* **Visibility Filtering:** Implemented scoped filtering for the travel table. **Unit Managers** can now see all entries for their unit, while **Standard Users** are restricted to entries matching their own `institutional_id`.
* **Frontend UX:** Updated `HeadcountMemberSelect.vue` to correctly handle loading states and display role-specific messages (e.g., directing standard users to contact their manager if not validated).

## Why is this needed?

1.  **Stale Data:** Users were reporting that updating the number of trips didn't change the CO2 total because the system prioritized the "stored" value from previous imports.
2.  **Access Issues:** Unit Managers were incorrectly restricted to seeing only their own travel records, hindering their ability to manage the unit's carbon report.
3.  **Authentication Errors:** The backend was using the wrong database field to verify roles, leading to "Permission Denied" errors for legitimate unit administrators.

## Type of change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📝 Documentation update
- [ ] 🎨 Design/UI improvement
- [ ] 🔧 Configuration change
- [x] 🧹 Code cleanup

## Related issues

- Closes #850
- Related to #

### DIFF
--- a/backend/app/api/v1/carbon_report_module.py
+++ b/backend/app/api/v1/carbon_report_module.py
@@ -123,15 +123,23 @@ def _has_global_or_principal_access_for_unit(
     current_user: User,
     unit: Optional[Unit],
 ) -> bool:
-    """Return whether the user has global or principal access for the unit."""
+    """Return whether the user has global or principal access for the unit.
+
+    Accred roles store the unit code in ``RoleScope.institutional_id``, matching
+    ``Unit.institutional_code``.  Test roles store ``Unit.institutional_id``
+    instead, so we check both fields.
+    """
     if any(isinstance(role.on, GlobalScope) for role in current_user.roles):
         return True
-    if unit is None or unit.institutional_id is None:
+    if unit is None:
         return False
-    return (
-        pick_role_for_institutional_id(current_user.roles, unit.institutional_id)
-        == RoleName.CO2_USER_PRINCIPAL
-    )
+    for field in (unit.institutional_code, unit.institutional_id):
+        if field is not None and (
+            pick_role_for_institutional_id(current_user.roles, field)
+            == RoleName.CO2_USER_PRINCIPAL
+        ):
+            return True
+    return False
 
 
 async def get_request_context(request: Request) -> dict:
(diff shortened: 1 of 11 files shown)


frontend/src/components/organisms/module/HeadcountMemberSelect.vue
```diff
@@ -72,18 +72,13 @@ interface SelectOption {
...
-const isNotValidated = computed(() => {
-  return (
-    members.value.length === 0 &&
-    !hasPermission(authStore.user?.permissions, 'modules.headcount', 'view')
-  );
-});
+const isUnitManager = computed(() =>
+  hasPermission(authStore.user?.permissions, 'modules.headcount', 'view'),
+);
+const isStandardUser = computed(() => !isUnitManager.value);
```
guilbep: @BenBotros MUST use permissions and not roles

guilbep (changes requested):

backend/app/api/v1/carbon_report_module.py
```diff
@@ -123,15 +123,23 @@ def _has_global_or_principal_access_for_unit(
...
     if any(isinstance(role.on, GlobalScope) for role in current_user.roles):
         return True
-    if unit is None or unit.institutional_id is None:
+    if unit is None:
         return False
-    return (
-        pick_role_for_institutional_id(current_user.roles, unit.institutional_id)
-        == RoleName.CO2_USER_PRINCIPAL
-    )
+    for field in (unit.institutional_code, unit.institutional_id):
```
guilbep: @BenBotros we use unit_institutional_id in the role now

guilbep (changes requested):

guilbep (approved):

Issue #835, opened 2026-04-13
[FEAT](Simulation) - Design DB Scheme


charlottegiseleweil:
## 🎯 Feature Overview 
Propose DB modification to include Simulation flow. 

## Worksession @guilbep & @charlottegiseleweil 20.04.2026

<img width="901" height="504" alt="Image" src="https://github.com/user-attachments/assets/836f10d1-d346-4d5c-96c9-3a97be1639b1" />

`percentage_of_last_year` is `OPTIONAL` and inside the` data-entries`
If present, we don't compute co2 as we do today, but by using the value of the precedent year

Proposed modification to DB scheme  : 
``` 
   carbon_reports {
    VARCHAR completion_progress
    INTEGER id PK
    INTEGER last_updated
    INTEGER overall_status
   INTEGER reference_year (optional)
    JSON stats
    INTEGER unit_id FK
    INTEGER year (optional)
    INTEGER carbon_project_id FK (optional)
  }
  carbon_projects {
    INTEGER id PK
    INTEGER unit_id FK
    INTEGER start_year
    INTEGER end_year
    VARCHAR name
    BOOLEAN is_public
  }
 
 
  carbon_projects ||--}o carbon_reports: "carbon_project_id"

```

Add a column : `Carbon_report_type` with 3 enum values : {`Calculator`, `Simulator_Explore`, ` Simulator_Plan` },

guilbep:
@charlottegiseleweil do we wait for domq before moving that in 'validated'?

charlottegiseleweil:
@guilbep Yes let's wait to get his input before we confirm :)

guilbep:
@charlottegiseleweil  pour l'explore, on prend quelle année pour les facteurs ? :p

guilbep:
- Must have 'year' for carbon_report (@domq dit que c'est fragile)
  -> Pour garder une forme normale, il faut que chaque carbon_report_type soit dans le carbon_project et que par conséquent, chaque carbon_report ait un carbon_project (même si defacto, seulement simulation_plan l'utilisera pleinement) 
 Corollaire => cela supprime le cas de la FK inexistante


nouveau schema:

```mermaidjs
erDiagram
carbon_reports {
    VARCHAR completion_progress
    INTEGER id PK
    INTEGER last_updated
    INTEGER overall_status
    INTEGER reference_year (optional)
    JSON stats
    INTEGER unit_id FK
    INTEGER year
[...]

PR #962, opened 2026-04-30
feat/856 simulation explore flow

### LINKED ISSUE #856: [TASK] Simulation / EXPLORE Flow

BenBotros:
## What does this change?

This PR introduces the **Simulator Explore flow** end-to-end, spanning database schema, backend services, and frontend UI:

- Adds a `carbon_projects` table to group carbon reports by unit, type (`Calculator`, `Simulator_Explore`, `Simulator_Plan`), and date range
- Adds a `carbon_report_type` enum column to `carbon_reports` and links every report to a project via `carbon_project_id`
- Implements a `GET /carbon-reports/simulator/explore/unit/{unit_id}/reference-year/{reference_year}/` endpoint that get-or-creates a Simulator Explore report with a 24h TTL, seeding it with research facility entries from the matching Calculator report
- Introduces a `SIMULATION_MODE` context variable (set via `X-Co2-Simulation: 1` request header) that bypasses module permission checks and validated-status filters for simulation pages
- Replaces the `emissionBreakdownRefreshSequence` pub/sub pattern with a direct `refreshEmissionBreakdownIfNeeded()` call after data mutations
- Builds out the `SimulationExplorePage` with per-module expansion panels, submodule data tables, a total CO₂ summary, and a download report card
- Adds a `collapsible` prop to `SubModuleSection` so submodules can render as flat cards instead of expansion items on the explore page

## Why is this needed?

Users need a lightweight, temporary workspace to explore the carbon impact of equipment, travel, purchases, research facilities, and lab activity **without committing changes to their main Calculator reports**. This feature provides that sandbox experience with automatic expiry (24h TTL) and pre-seeded data from the reference year.

## Type of change

- [x] ✨ New feature
- [x] 🎨 Design/UI improvement
- [x] 🧹 Code cleanup

## Related issues

- Closes #856
- Related to #


frontend/src/pages/app/SimulationExplorePage.vue
```diff
@@ -1,7 +1,220 @@
...
+const defaultThreshold: Threshold = {
+  type: MODULES_THRESHOLD_TYPES[0],
+  value: 0,
+};
+
+const mountPrimaryCharts = ref(false);
+
+const modules = computed(() => {
+  return MODULES_ORDER.filter((type) => type !== MODULES.ResearchFacilities)
+    .map((type) => {
```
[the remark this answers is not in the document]
BenBotros: i know

guilbep:
# Code Review

## Overview
OK WRONG: cf https://github.com/EPFL-ENAC/co2-calculator/issues/835#issuecomment-4351220055
Introduces the **Simulator Explore** flow end-to-end: a `carbon_projects` table groups reports by type (`Calculator` / `Simulator_Explore` / `Simulator_Plan`), every report gets a nullable `carbon_project_id`, and a new `GET /carbon-reports/simulator/explore/unit/{unit_id}/reference-year/{reference_year}/` get-or-creates a 24h-TTL Explore report seeded with research-facility data from the matching Calculator report. (if the code is dead, remove it), do a remove dead-code of the PR) A new `X-Co2-Simulation: 1` request header (set client-side when the URL contains `/simulation/`) routes module queries to the Explore report instead of the Calculator one.
-- TODO: remove X-CO2-Simulation, and replace by query-param (?carbon_project_type=0,1,2) WE MUST BE sure the user cannot circumvent the permissions
42 files / +1504 / −195. 5/10 PR checks failing.

---

## 🟥 Blocking — must fix before merge

### B1. `bulk_upsert` will crash at runtime: dropped unique constraint, no replacement
The migration drops `uq_carbon_reports_unit_year` and the model's `__table_args__` removes the constraint, but `CarbonReportRepository.bulk_upsert` still calls:
```python
.on_conflict_do_nothing(index_elements=["unit_id", "year"])
```
PostgreSQL requires a unique index/constraint matching the `ON CONFLICT` target — the upgrade adds none. First call after migration will raise *"there is no unique or exclusion constraint matching the ON CONFLICT specification"*. Either:
- Re-add a partial unique index (e.g. unique on `(unit_id, year)` `WHERE carbon_project_id IS NOT NULL` or scoped to Calculator reports), or
- Rewrite `bulk_upsert` to query-then-insert.

This must be tested in CI against a real Postgres, not just SQLite.

### B2. Header-controlled `validated_only` flag exposes unvalidated emissions
`backend/app/api/v1/carbon_report_module_stats.py` (`get_validated_totals`):
```python
is_simulator = request.headers.get("X-Co2-Simulation") == "1"
emission_stats = await DataEntryEmissionService(db).get_stats_by_carbon_report_id(
    carbon_report_id=carbon_report_id,
    validated_only=not is_simulator,
)
```
Any authenticated user can pass an arbitrary Calculator `carbon_report_id` plus `X-Co2-Simulation: 1` and read **unvalidated** module data that wouldn't otherwise be visible. The author already implemented the correct pattern two functions over in `get_emission_breakdown` — derive simulator-ness from `CarbonProject.carbon_report_type` for the given `carbon_report_id`. Apply the same fix here.

### B3. Inner-join to `CarbonProject` makes pre-migration reports invisible
`carbon_report_repo.py` rewrote both `list_by_unit` and `get_by_unit_and_year` to `JOIN CarbonProject` (inner). `carbon_project_id` is nullable and the migration has **no backfill**. After deploy, any existing `carbon_reports` row with `carbon_project_id IS NULL` disappears from these queries — i.e. every report on dev/prod/staging. Add a data-migration step:
```sql
INSERT INTO carbon_projects (unit_id, carbon_report_type, ...) SELECT DISTINCT ...;
UPDATE carbon_reports SET carbon_project_id = ...;
ALTER TABLE carbon_reports ALTER COLUMN carbon_project_id SET NOT NULL;
```
or relax the join semantics.

---

## 🟧 Should fix before merge

### S1. Permission bypass via simulator seeding
`get_or_create_simulator_explore_carbon_report` only calls `_require_unit_access` — there's no module-level permission check. `_seed_research_facility_entries` then copies all research-facility `DataEntry` rows from the Calculator report into the Explore report. A unit member without `modules.research_facilities` permission can read that data through the Explore endpoints that follow. Either gate the new endpoint on the relevant `modules.*` permissions, or skip seeding modules the user can't view.

### S2. TTL deletion is destructive on read + race-prone



--> MAYBE USE the data_ingestion_job and poller instead 
`get_or_create_explore` calls `self.delete(existing.id)` whenever the report is older than `ttl_seconds`. Two simultaneous post-expiry reads will both:
1. See `existing` as expired,
2. Both delete (one is a no-op),
3. Both create a fresh report — leaving two rows for the same `(unit_id, reference_year)` Explore.

`get_explore_by_unit_and_reference_year` masks this by `LIMIT 1`, so duplicates accumulate silently. Add a unique constraint on `(carbon_project_id, year)` for Explore projects, or wrap in a `SELECT … FOR UPDATE` / advisory lock. Also worth documenting that any unsaved Explore work disappears at the 24h boundary on the next read.

### S3. Cascade on `delete_by_report` may leave orphan `data_entry_emissions`
The new `delete_by_report` directly issues a SQL `DELETE FROM data_entry` (bypassing ORM cascades). If `data_entry_emissions.data_entry_id` doesn't have `ON DELETE CASCADE` at the DB level, this raises an FK violation. Worth checking the FK definition; either rely on a real cascade or also delete emissions explicitly first.

### S4. `_get_percentage_override_kg` is unrelated scope
A ~80-line addition to `data_entry_emission_service.py` introduces percentage-of-prior-year emission resolution (matching by `user_institutional_id`/`name`, fetching previous-year report/module/entry). This is a Plan-flow feature, not part of "Simulator Explore," and isn't mentioned in the PR description. It deserves its own PR with focused tests on the matching heuristics. Keeping it here makes the diff harder to revert if Explore needs to roll back.

---

## 🟨 Nits & follow-ups

- **Frontend trusts the URL globally.** `frontend/src/api/http.ts` sets `X-Co2-Simulation: 1` on **every** ky request whenever `location.pathname.includes('/simulation/')`. Worth a comment audit that no shared service called from a simulation page hits a Calculator-only endpoint — otherwise that endpoint silently flips behavior.
- **Dead `collapsible=false` branch in `SubModuleSection.vue`.** Both callers (`ModuleTableSection.vue` and `SimulationExplorePage.vue`) use the default `collapsible=true`. Either drop the `q-card` else-if branch or wire it up.
- **`_require_unit_access` duplicates policy logic.** It re-implements role-scope walking that `policy.py::_evaluate_data_filter_policy` already does. Consider lifting to a shared helper to keep the data-boundary check in one place — and to ensure the simulator endpoint and existing endpoints stay in lockstep.
- **`headcountMemberCount` re-mount key.** `:key="headcountMemberCount"` on `<HeadcountMemberSelect>` re-mounts the whole component on every count change, refetching members and dropping any in-progress user input. Cleaner: expose `fetchMembers` via `defineExpose` and call it from the parent on the relevant lifecycle.
- **`useWorkspaceStore()` resolved at `useModuleStore` setup.** Pinia handles cross-store imports, but moving the call inside `refreshEmissionBreakdownIfNeeded` would make the dependency graph less surprising and ease testing.
- **`CarbonReport.unit_id` typing changed `string → number`.** Correct, but verify no downstream consumer (charts, formatters) was string-coercing it implicitly.
- **`is_viewable_by_unit_members` field on `CarbonProject`.** Dead-stored: never read in this PR. Either wire it or drop it from the migration to keep the schema honest.
- **Migration filename / id quirk.** `2026_05_04_0945-05d68c9a6054_add_simulation_carbon_reports.py` overlaps with the recent migrations consolidation. Confirm no other branch is producing a competing head — Alembic will refuse to merge.

---

## Test coverage

- `tests/conftest.py` and the repo tests are correctly updated to seed a `CarbonProject` before creating a `CarbonReport` — good.
- **Missing:** no test exercises the new `/simulator/explore/...` endpoint end-to-end (TTL expiry, seeding, idempotency).
- **Missing:** no test for the Calculator/Explore inner-join cutoff (B3).
- **Missing:** no test asserting `bulk_upsert` still works after the migration (B1).
- The three new `_get_report_for_data_entry` patches in `test_data_entry_emission_service.py` are required only because the service was changed to fall through; that's fine but flags how invisibly-coupled `prepare_create` has become to the new code path.

---

## Summary

The architectural shape (typed projects + TTL'd Explore reports) is sound and the server-side check in `get_emission_breakdown` shows the author understands the trust boundary. But three issues will cause real damage on merge: a guaranteed runtime crash in `bulk_upsert`, a header-trust authorization gap on `get_validated_totals`, and a silent data-visibility regression for all pre-existing reports. Plus an unrelated 80-line feature that should split off. With those addressed, the rest is solid feature work.

backend/app/models/carbon_report.py
```diff
@@ -69,11 +96,6 @@ class CarbonReport(CarbonReportBase, table=True):
     __tablename__ = "carbon_reports"
     id: Optional[int] = Field(default=None, primary_key=True)
 
-    # Unique constraint for (year, unit_id)
-    __table_args__ = (
-        UniqueConstraint("unit_id", "year", name="uq_carbon_reports_unit_year"),
```
guilbep, line 74: We need to create a new constraint only for the carbon_project type (co2 calculator)
which is uniq across co2 calculator
I'm not sure how to do it without having to 'denormalize' the carbon_project type (which mean having the type in both carbon_project and carbon_report)

but we need to find a way to avoid having more than one carbon_report per year for the carbon_report


AT the latest. We need to have the constraint at the carbon_project level just for this type, and maybe that' ll be enough

backend/app/services/carbon_report_service.py
```diff
@@ -28,24 +35,163 @@ def __init__(self, session: AsyncSession):
...
+        self,
+        *,
+        unit_id: int,
+        reference_year: int,
+        ttl_seconds: int = 24 * 60 * 60,
+    ) -> CarbonReportRead:
+        """Get-or-create a Simulator Explore report with a TTL.
+
+        The explore report uses ``year = reference_year`` (year is always non-null).
+        If an existing report is older than ``ttl_seconds`` it is deleted and a new
```
guilbep: get should be idempotent. no get_or_create at the same time.
for the ttl_seconds, we should probably have a background -task or we should defer to a new issue in the future

guilbep (changes requested):

backend/alembic/versions/2026_05_04_0952-3f8147b5e516_search_locations.py
```diff
@@ -21,7 +21,7 @@
 
 # revision identifiers, used by Alembic.
 revision: str = "3f8147b5e516"  # noqa: F841
-down_revision: Union[str, Sequence[str], None] = "c2d4e6f8a012"  # noqa: F841
+down_revision: Union[str, Sequence[str], None] = "05d68c9a6054"  # noqa: F841
```
ymarcon: Not supposed to modify the down version of a former migration script.
The new migration script comes after the existing ones.

backend/alembic/versions/2026_05_04_0945-05d68c9a6054_add_simulation_carbon_reports.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+# revision identifiers, used by Alembic.
+revision: str = "05d68c9a6054"  # noqa: F841
+down_revision: Union[str, Sequence[str], None] = "c2d4e6f8a012"  # noqa: F841
+branch_labels: Union[str, Sequence[str], None] = None  # noqa: F841
+depends_on: Union[str, Sequence[str], None] = None  # noqa: F841
+
+
+def upgrade() -> None:
```
ymarcon, line 32: Looks like this migration scripts has a lot of issues. If it was implemented by a LLM, drop it and recreate it properly with Pydantic/SQLModel.

backend/app/api/v1/carbon_report.py
```diff
@@ -17,6 +21,25 @@
 from app.services.carbon_report_module_service import CarbonReportModuleService
 from app.services.carbon_report_service import CarbonReportService
 
+_EXPLORE_TTL_SECONDS = 24 * 60 * 60  # 24 hours
+
+
+async def _refresh_explore_background(
```
ymarcon, line 27: That is not a proper garbage collector task: this should be triggered on a regular basis, either by a cronjob (recommended) or by a global background task (if you can make sure it is not running in all deployed instances of the app)

ymarcon (changes requested):

PR #1105, opened 2026-05-08
feat/874: reject legacy kg_co2eq override in DataEntry.data

### LINKED ISSUE #874: [REFACTOR](Travel): Remove travel's kgco2eq from  data entries
## 🎯 Task Overview (Why?)

CSV imports store the source system's pre-computed CO2 value in `DataEntry.data["kg_co2eq"]`. `DataEntryEmissionService.prepare_create()` treats this as an override and skips the emission formula entirely. When a user later edits the entry (e.g. changes `number_of_trips`), the stale value persists and CO2 is never recalculated.

A temporary fix was added in #867 that strips `kg_co2eq` from `data` before recomputing on update — this treats the symptom, not the cause.

`data_entry.data` should be a clean bag of raw input fields. Computed/overridden emission values belong in `data_entry_emissions`.

## Propsed Fix

1. Stop writing `kg_co2eq` into `DataEntry.data` during CSV import — remove it from `data_payload` in `professional_travel_api_provider.py` and any other providers doing the same.
2. Write the CSV override directly into `DataEntryEmission.kg_co2eq` after creating the `DataEntry`, bypassing `prepare_create` for entries with a trusted source value.
3. Remove the stripping hack from `DataEntryEmissionService.update_emission()`.
4. Add a schema-level guard so `kg_co2eq` is rejected in `DataEntry.data`, preventing recurrence.

## ✅ Success Criteria
- [ ] kg_co2eq` no longer appears in any `data_entry.data` payload
- [ ] Editing a CSV-imported entry (e.g. `number_of_trips`) correctly recomputes CO2
- [ ] Stripping hack removed from `update_emission()

BenBotros:
## What does this change?

Removes the `__kg_co2eq_override__` carrier key from `DataEntry.data`
and replaces the pattern with a direct write into `DataEntryEmission` at
import time. Concretely:

- **Schema guard** (`data_entry.py`): Adds a `@model_validator` to both
  `DataEntryCreate` and `DataEntryUpdate` that rejects any payload
  containing `kg_co2eq` or `__kg_co2eq_override__` in `data`, preventing
  the legacy pattern from re-emerging via the API.
- **Emission service** (`data_entry_emission_service.py`): Removes the
  B-H1 fallback that read `__kg_co2eq_override__` from `data_entry.data`
  when no function-arg override was passed. `upsert_by_data_entry` and
  the async recalc path now always formula-recompute.
- **Travel API provider** (`professional_travel_api_provider.py`):
  Removes the `KG_CO2EQ_OVERRIDE_KEY` write. Instead, after bulk-insert,
  resolves emission types and writes `DataEntryEmission` rows directly
  with `kg_co2eq=override` and `source="import_override"`.
- **Base CSV provider** (`base_csv_provider.py`): Same — removes the
  `KG_CO2EQ_OVERRIDE_KEY` write and adds a direct `DataEntryEmission`
  bulk-create for overrides in the async path.
- **Tests**: Updates both unit and integration tests to assert that
  `__kg_co2eq_override__` is absent from `DataEntry.data`, and rewrites
  the integration test suite to reflect the new contract:
  - Import writes override emissions directly → value is present at rest.
  - `upsert_by_data_entry` (user edits, async recalc) always
    formula-recomputes → import override value does not survive recalc.

## Why is this needed?

The `__kg_co2eq_override__` carrier was a workaround introduced so that
source-provided `kg_co2eq` values (Tableau's `OUT_CO2_CORRECTED`, CSV
columns) would survive the async recalc hop under `BULK_PATH_PURE_ASYNC`.
It had two problems: it polluted `DataEntry.data` with an internal
implementation detail, and it caused every subsequent recalc (including
user edits) to silently preserve the import value instead of
formula-recomputing. Writing the override directly into
`DataEntryEmission` at import time is the correct separation — the data
entry stays clean, and recalcs always use the formula.

## Type of change

- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [x] 🧹 Code cleanup

## Code quality checklist

- [ ] Documentation updated for new features

## Testing checklist

## Related issues

- Closes #874

### DIFF
--- a/backend/alembic/versions/2026_05_13_1008-0cb07d36bdf0_add_kgco2_override_to_data_entry.py
+++ b/backend/alembic/versions/2026_05_13_1008-0cb07d36bdf0_add_kgco2_override_to_data_entry.py
@@ -1,0 +1,49 @@
+# codeql[py/unused-global-variable]
+"""add kgco2 override to data entry
+
+Revision ID: 0cb07d36bdf0
+Revises: a7b2f8c1d3e6
+Create Date: 2026-05-13 10:08:00.951062
+
+"""
+
+from typing import Sequence, Union
+
+import sqlalchemy as sa
+
+from alembic import op
+
+__all__ = [
+    "revision",
+    "down_revision",
+    "branch_labels",
+    "depends_on",
+]
+
+
+# revision identifiers, used by Alembic.
+revision: str = "0cb07d36bdf0"  # noqa: F841
+down_revision: Union[str, Sequence[str], None] = "a7b2f8c1d3e6"  # noqa: F841
+branch_labels: Union[str, Sequence[str], None] = None  # noqa: F841
+depends_on: Union[str, Sequence[str], None] = None  # noqa: F841
+
+
+def upgrade() -> None:
+    """Upgrade schema."""
+    op.add_column(
+        "data_entries", sa.Column("kg_co2eq_override", sa.Float(), nullable=True)
+    )
+    op.create_index(
+        "ix_data_entries_kg_co2eq_override_notnull",
+        "data_entries",
+        ["kg_co2eq_override"],
+        postgresql_where="kg_co2eq_override IS NOT NULL",
+    )
+
+
+def downgrade() -> None:
+    """Downgrade schema."""
+    op.drop_index(
+        "ix_data_entries_kg_co2eq_override_notnull", table_name="data_entries"
+    )
+    op.drop_column("data_entries", "kg_co2eq_override")
--- a/backend/app/models/data_entry.py
+++ b/backend/app/models/data_entry.py
@@ -4,7 +4,7 @@ from datetime import datetime
 from enum import Enum
 from typing import Optional
 
-from sqlalchemy import Column, DateTime, Integer
+from sqlalchemy import Column, DateTime, Float, Integer
 from sqlmodel import JSON, Field, SQLModel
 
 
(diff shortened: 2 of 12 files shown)


guilbep:
1) kg_co2eq has been replaced by __kg_co2eq_override__
2) maybe the solution is

Suggestion for follow-up tech debt issue:

Promote `data['__kg_co2eq_override__']` to a first-class nullable column on `DataEntry` (e.g. `kg_co2eq_override FLOAT NULL`) instead of keeping it only inside the JSON payload.

The important point is that the override belongs to the `DataEntry` itself, not to the computed `DataEntryEmission`.

Rationale:
in case we need to recompute/regenerate `DataEntryEmission` rows later (for example after uploading new emission factors), we must still preserve manual CO2 overrides.

Newly generated `DataEntryEmission` rows should reuse the override value instead of recomputing emissions normally. Because of that, the override must live on the `DataEntry` source of truth, and we need to efficiently filter/select overridden entries during bulk recompute jobs.

Benefits:

* explicit typed schema instead of magic JSON key
* simpler ORM/query logic
* easier filtering/indexing for bulk recompute jobs
* clearer ownership of the override state

No separate boolean flag should be needed since override presence can be inferred with:

```sql
kg_co2eq_override IS NOT NULL
```

guilbep:
not necessary anymore. We use __kgco2_eq_overide__

PR #1108, opened 2026-05-08
feat(867): wire 'Open year for users' button + reflect is_started in UI

guilbep:
## Summary

Unit 5 of issue #867 (back-office data-management consolidation).

The "Open year for users" button on the back-office data-management page rendered but did nothing — there was no `@click` handler. This wires it up:

- New `openForUsers(year)` action on the `yearConfig` store — thin wrapper over `updateConfig` that PATCHes `{ is_started: true }`.
- Button now has `@click="handleOpenForUsers"` showing a positive toast on success and a negative toast on error (mirrors `handleCreateYear`).
- Button is `:disable`d when `anyModuleIncomplete` (existing) **or** the year is already open (new). Tooltip text differs per reason.
- New `q-chip` near the year selector at-a-glance shows whether the year is open to users (`Open to users` / `Not yet open`) — green chip with `lock_open` when open, neutral with `lock` when not.
- New i18n keys (en/fr): `data_management_year_already_open`, `data_management_year_opened_success`, `data_management_year_is_open`, `data_management_year_is_not_open`.

Distinct from "Create year" (U1's territory) — does not touch that flow.

## Independence from U1

If U1 ships first, new years will already have `is_started=true` so this button is automatically disabled with the "already open" tooltip — desired behavior. If U1 has not shipped yet, this still works for legacy years that have `is_started: false` (e.g. 2025).

## Test plan

- [x] `vue-tsc --noEmit` passes.
- [x] `eslint .` passes.
- [x] `prettier --check` passes.
- [x] New Playwright integration spec at `frontend/tests/integration/data-management.spec.ts` lists 2 cases (button enabled + PATCH on click; button disabled when year already open). Spec stubs the year-configuration API and skips cleanly if the back-office route is unreachable in the test env.
- [ ] Full Playwright e2e run not executed locally — no built `dist/spa` and the back-office route requires auth; CI is the right place to run this.
- [ ] Manual smoke: verify clicking the button on a 2025 year flips the chip to "Open to users" and disables the button.

## Notes

- The unit instructions called for vitest unit tests for the store action, but this project does not use vitest (Playwright only). Skipped accordingly.

### DIFF
--- a/frontend/src/i18n/backoffice_data_management.ts
+++ b/frontend/src/i18n/backoffice_data_management.ts
@@ -632,4 +632,20 @@ export default {
     en: 'All mandatory factor and reference uploads must be completed before opening the year for users.',
     fr: "Tous les téléversements obligatoires de facteurs et de références doivent être complétés avant d'ouvrir l'année pour les utilisateurs.",
   },
+  data_management_year_already_open: {
+    en: 'Year is already open to users',
+    fr: "L'année est déjà ouverte aux utilisateurs",
+  },
+  data_management_year_opened_success: {
+    en: 'Year {year} opened to users',
+    fr: 'Année {year} ouverte aux utilisateurs avec succès',
+  },
+  data_management_year_is_open: {
+    en: 'Open to users',
+    fr: 'Ouverte aux utilisateurs',
+  },
+  data_management_year_is_not_open: {
+    en: 'Not yet open',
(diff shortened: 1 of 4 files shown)


guilbep:
Rolled into #1107 (umbrella feat/857 → dev). Head commit 34ffedf7 is reachable from feat/857; closing to clear the review queue.

PR #1112, opened 2026-05-08
feat/892: join Location table for train sort fields and enrich response

### LINKED ISSUE #892: [BUG](Travel) Table : sort functionality not working on `Names`
"Sort" feature on names not functional 

 _Originally posted by @charlottegiseleweil in [#850](https://github.com/EPFL-ENAC/co2-calculator/issues/850#issuecomment-4278572508)_``

BenBotros:
## What does this change?

Joins the `Location` table (via aliased `OriginLocation` / `DestLocation`) into the train data-entry query so that `origin_name` and `destination_name` are resolved from the database rather than read raw from `DataEntry.data`. Adds `origin_name`, `destination_name`, and `distance_km` to the sort map for train entries, with `distance_km` using `COALESCE(DataEntryEmission.additional_value, DataEntry.data["distance_km"])` so the computed emission value always takes precedence. Also exposes `origin_iata`, `destination_iata`, `origin_name`, `destination_name`, and `distance_km` in the base handler's response field map.

## Why is this needed?

`origin_name` and `destination_name` were previously read directly from `DataEntry.data`, which meant sorting by those fields was unreliable and the values could diverge from the canonical `Location` records. The `distance_km` sort similarly bypassed the authoritative emission value stored in `DataEntryEmission.additional_value`.

## Type of change

- [x] 🐛 Bug fix
- [x] ✨ New feature

## Code quality checklist

## Testing checklist

Two integration tests added in `test_sort_uses_location.py`:
1. Verifies that sorting by `origin_name` uses `Location.name` from the JOIN — entries are returned A→Z even when Location rows were inserted in reverse order
2. Verifies that `distance_km` sort uses `COALESCE(additional_value, data["distance_km"])`, so `DataEntryEmission.additional_value` wins over the raw data field

## Related issues

- Closes #
- Related to #1124

### DIFF
--- a/backend/app/modules/professional_travel/schemas.py
+++ b/backend/app/modules/professional_travel/schemas.py
@@ -104,6 +104,8 @@ class ProfessionalTravelPlaneHandlerResponse(DepartureDateMixin, DataEntryRespon
     number_of_trips: int = 1
     origin: Optional[str] = None
     destination: Optional[str] = None
+    origin_name: Optional[str] = None
+    destination_name: Optional[str] = None
     distance_km: Optional[float] = None
     note: Optional[str] = None
     kg_co2eq: Optional[float] = None
@@ -206,6 +208,9 @@ class ProfessionalTravelBaseModuleHandler(BaseModuleHandler):
         "number_of_trips": DataEntry.data["number_of_trips"].as_float(),
         "traveler_name": MemberEntry.data["name"].as_string(),
         "kg_co2eq": DataEntryEmission.kg_co2eq,
+        # Plane destination
+        "origin_iata": DataEntry.data["origin_iata"].as_string(),
+        "destination_iata": DataEntry.data["destination_iata"].as_string(),
     }
 
     def to_response(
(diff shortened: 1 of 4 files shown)


guilbep:
@BenBotros  we SHOULD not do that, cf copilot commit.
1) you should join with he location table to get the origin_name, destination_name, it should not be part of the data_entry EVER.

- Add integration test to confirm that please


        "origin_name": DataEntry.data["origin_name"].as_string(),
        "destination_name": DataEntry.data["destination_name"].as_string(),



- for distance_km is in data_entry_emission in additional_value, not in DATAENTRY.dat
        "distance_km": DataEntry.data["distance_km"].as_float(),
        
        
        
        
        
        cf distance_km sorting is currently based on DataEntry.data["distance_km"], but travel listings populate the displayed distance_km from DataEntryEmission.additional_value (source of truth) when present (see DataEntryRepository enrichment). This will make table sorting by distance appear broken/inconsistent for rows where data.distance_km is null but additional_value is set. Consider sorting by coalesce(DataEntryEmission.additional_value, DataEntry.data["distance_km"].as_float()) so ordering matches what the API returns in distance_km.
        
        
        
        
        
        @ymarcon what do you think ?

guilbep (changes requested):
please don't store in data_entry information that don't come from the user

guilbep (approved):

PR #1126, opened 2026-05-11
feat/1124 disambiguate locations by country for train module

### LINKED ISSUE #1124: [BUG](Travel) Error 500 : `{"detail":"Failed to create data entry"}` when adding a train
Image

## How to reproduce 
Seems to fail only when BERNE selected. !!!

## Success criteria

- [ ] Fix the bug : we can create data entries adding trains
- [ ] Create a test to ensure it doesnt happen again . `{"detail":"Failed to create data entry"}`

BenBotros:
## What does this change?

Adds a unique constraint `uq_location_name_mode_country` on `(name, transport_mode, country_code)` in the `locations` table, and threads `origin_country_code` / `destination_country_code` through the entire stack — train schemas, location repository, location service, `ModuleForm.vue`, and the module store — so that location lookups can disambiguate train stations that share the same name but belong to different countries. Also fixes a date input reset bug in `ModuleForm` via a reactive key.

## Why is this needed?

When two train stations share the same name (e.g. "Berne" in CH and DE), `get_by_name()` called `scalar_one_or_none()` on a query returning multiple rows, raising `sqlalchemy.exc.MultipleResultsFound` and returning a 500 error. By passing the country code from the frontend through to the database query, the lookup is unambiguous.

## Type of change

- [x] 🐛 Bug fix

## Code quality checklist

## Testing checklist

Two integration tests added in `test_duplicate_station_name.py`:
1. Verifies `pre_compute` correctly resolves a station when two rows share the same name but differ by `country_code`
2. Documents the old failure mode — without a country code, `MultipleResultsFound` is raised (confirming the fix is required on the caller side)

## Related issues

- Closes #1124


ymarcon (approved):

guilbep (changes requested):
All copilot comments seemed valid
Special interest on the 'country' won't be saved during patch because the frontend don't have it

- integrations tests should test create AND get -> then patch

guilbep (changes requested):
The uniqueness of Berne/DE won't. work since we have 'two' Berne train station with different gps coordinates in the database

TL;DR: ADD natural_key column in DB, it's not part of the csv. It is a database-derived field used during ingestion
```raw
CSV row
   ↓
Backend computes natural_key
   ↓
UPSERT on natural_key
   ↓
DB stores/updates row
```

Best solution:
- Add a stable unique station identifier
- Frontend sends identifier
- Backend stores identifier
- Name/country are display metadata only
- Do NOT use coordinates as identity
- Do NOT use (name, country) as uniqueness constraint


# 📍 Location Identity Strategy (CSV Import + Database)

## 🎯 Goal

We need a stable and idempotent way to manage locations (train stations + airports) across repeated CSV imports.

---

## 🧠 Core Principles

### 1. Database `id`

* Type: `INT` (auto-increment)
* Purpose: internal database primary key
* Never exposed as external identity
* Must never change once created

---

### 2. `natural_key` (IMPORTANT)

* Type: `STRING`
* Purpose: deduplication + UPSERT key during CSV imports
* Must be deterministic and stable across re-imports
* Must uniquely identify a logical location

---

## 🚫 What NOT to do

Do NOT use as identity:

* Only `name` (ambiguous duplicates exist)
* Only `coordinates` (can drift slightly between datasets)
* Only `name + country` (still ambiguous in real data)

⚠️ Coordinates are NOT a primary identifier.
They are only used as a fallback disambiguation signal for trains.

---

## ✅ Identity Rules

### ✈️ Airports (plane mode)

If `iata_code` exists:

```
natural_key = "plane:" + iata_code
```

Example:

```
plane:NIM
```

✔ Stable because IATA is globally unique and standardized

---

### 🚆 Train stations

No global standard ID exists, so we build a composite key:

```
natural_key = "train:" + country_code + ":" + normalized_name + ":" + latitude + ":" + longitude
```

Normalization rules for `name`:

* lowercase
* trimmed
* collapse multiple spaces

Example:

```
train:ch:berne:46.94864:7.43654
```

⚠️ Coordinates are included ONLY because trains lack a universal identifier.
If a proper railway ID becomes available later, it should replace coordinates in this key.

---

## 🔁 Import Behavior (CRITICAL)

CSV imports must be idempotent using UPSERT:

```sql
INSERT INTO locations (...)
VALUES (...)
ON CONFLICT (natural_key)
DO UPDATE SET
  name = EXCLUDED.name,
  latitude = EXCLUDED.latitude,
  longitude = EXCLUDED.longitude,
  country_code = EXCLUDED.country_code,
  iata_code = EXCLUDED.iata_code;
```

---

## 🧩 Summary

* `id` → internal DB identity (INT, immutable)
* `natural_key` → deterministic identity for deduplication
* `iata_code` → external airport identifier (when available)
* coordinates → fallback disambiguation only (not identity)

---

## 🧠 Mental Model

* If a real-world standard exists (IATA) → use it
* If no standard exists → build deterministic composite key
* Always ensure re-import produces the same `natural_key`

backend/alembic/versions/2026_05_12_1116-8f29ed82872b_add_natural_key_to_locations.py
```diff
@@ -0,0 +1,144 @@
...
+]
+
+
+# revision identifiers, used by Alembic.
+revision: str = "8f29ed82872b"  # noqa: F841
+down_revision: Union[str, Sequence[str], None] = "1e56d303cc85"  # noqa: F841
+branch_labels: Union[str, Sequence[str], None] = None  # noqa: F841
+depends_on: Union[str, Sequence[str], None] = None  # noqa: F841
+
+_AGG_WHERE = (
```
guilbep, line 31: @BenBotros  Si on mets les index ailleurs, ce sera plus facile de refactor la migration! 


* Don’t try to make Alembic autogenerate complex PostgreSQL partial indexes.
* Keep them hand-written in migrations.
* Use readable SQL predicates (`IN (...)`) instead of PostgreSQL reflected forms (`ANY(ARRAY[...])`).
* Store predicates as shared constants to avoid duplication.
* Treat these indexes as DB/performance concerns, not ORM concerns.

Recommended pattern:

```python id="x4y0q8"
# indexes.py

AGG_WHERE = """
job_type = 'aggregation'
AND state IN ('NOT_STARTED', 'QUEUED', 'RUNNING')
"""

RECALC_WHERE = """
job_type = 'emission_recalc'
AND state IN ('NOT_STARTED', 'QUEUED', 'RUNNING')
AND module_type_id IS NOT NULL
AND data_entry_type_id IS NOT NULL
AND year IS NOT NULL
"""
```

```python id="op0nq2"
# alembic migration

op.create_index(
    "ix_ingestion_aggregation",
    "ingestion",
    ["organization_id"],
    postgresql_where=sa.text(AGG_WHERE),
)
```

Reason:
Alembic/SQLAlchemy cannot reliably round-trip advanced PostgreSQL predicates and will create migration noise/churn.

backend/app/modules/professional_travel/schemas.py
```diff
@@ -148,6 +149,10 @@ class ProfessionalTravelTrainHandlerCreate(
     user_institutional_id: str
     origin_name: str
     destination_name: str
+    origin_natural_key: Optional[str] = None
+    destination_natural_key: Optional[str] = None
+    origin_country_code: Optional[str] = None
```
guilbep: ok my intuition is that we should maybe store the 'natural_key' instead of country, because if ww have origin_name + origin_country we'll have the same problem for ben

Instinct tell me that we should change the drop-down select object to contain, country, name and natural_key

backend/app/modules/professional_travel/schemas.py
```diff
@@ -302,13 +311,38 @@ async def pre_compute(self, data_entry: Any, session: Any) -> dict:
...
         if origin_name is None or destination_name is None:
             return {}
 
         loc_service = LocationService(session)
-        origin = await loc_service.get_location_by_name(origin_name)
-        dest = await loc_service.get_location_by_name(destination_name)
+
+        # Prefer natural_key (unique — never ambiguous); fall back to name+country_code
+        # for entries created before natural_key was stored on the data entry.
+        if origin_natural_key:
```
guilbep: NO FALLBACK

backend/app/services/location_service.py
```diff
@@ -69,17 +69,42 @@ async def search_locations(
         # Convert to DTOs
         return [LocationRead.model_validate(location) for location in locations]
 
-    async def get_location_by_name(self, name: str) -> Optional[Location]:
+    async def get_location_by_name(
```
guilbep: @BenBotros  get_location_by_name dead code now ! ?

backend/app/repositories/location_repo.py
```diff
@@ -156,9 +157,41 @@ async def get_by_id(self, location_id: int) -> Optional[Location]:
         result = await self.session.get(Location, location_id)
         return result
 
-    async def get_by_name(self, name: str) -> Optional[Location]:
-        """Get location by name."""
-        statement = select(Location).where(col(Location.name) == name)
+    async def get_by_name(
+        self,
```
guilbep: @BenBotros  pretty it's dead code now?

guilbep (changes requested):

guilbep (approved):

PR #1214, opened 2026-05-18
fix: bring back module chart toggle

BenBotros:
## What does this change?
Removes the `isModuleValidated` guard that was hiding module charts for non-validated modules, and fixes the `showControls` prop default so chart view toggle controls are visible by default.

## Why is this needed?
Module charts were not showing for modules that hadn't been explicitly validated, even when results existed. The `showControls` prop also lacked a default value, causing the breakdown/type view toggle to be hidden unintentionally. The original intent was to hide controls only in print mode or when `forcedView` is set.

## Type of change
- [x] 🐛 Bug fix

## Code quality checklist
- [x] I confirm that my contribution is original and that I assign all intellectual property rights in this contribution to EPFL, retaining no ownership rights.

## Testing checklist

## Related issues
- Closes #

### DIFF
--- a/frontend/src/components/organisms/module/ModuleCharts.vue
+++ b/frontend/src/components/organisms/module/ModuleCharts.vue
@@ -132,13 +132,19 @@ import {
 } from 'src/constant/charts';
 import { getEmissionTypeBreakdownInfoKey } from 'src/constant/emissionTypeBreakdownInfo';
 
-const props = defineProps<{
-  type: Module;
-  showEvolutionChart?: boolean;
-  forcedView?: 'breakdown' | 'type';
-  showControls?: boolean;
-  printMode?: boolean;
-}>();
+const props = withDefaults(
+  defineProps<{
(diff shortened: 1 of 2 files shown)


frontend/src/components/organisms/module/ModuleCharts.vue
```diff
@@ -132,13 +132,16 @@ import {
...
-  showEvolutionChart?: boolean;
-  forcedView?: 'breakdown' | 'type';
-  showControls?: boolean;
-  printMode?: boolean;
-}>();
+const props = withDefaults(
+  defineProps<{
+    type: Module;
+    showEvolutionChart?: boolean;
+    forcedView?: 'breakdown' | 'type';
```
guilbep, line 139: @BenBotros not gonna merge because of that warning!

guilbep (changes requested):

guilbep (approved):

Issue #1219, opened 2026-05-18
[BUG](BackOffice Configuration) Equipments Data Upload Fail


charlottegiseleweil:
importing [equipments_data.csv](https://epflch.sharepoint.com/:x:/r/sites/VPS-Environmental_Data-CO2-calculator-data/Documents%20partages/SEED_DATA/equipments_data.csv?d=wcb491c70f7ef4846b331086a0101cebb&csf=1&web=1&e=MNvZU5) in stage 

<img width="1240" height="500" alt="Image" src="https://github.com/user-attachments/assets/6c1464da-128d-4adc-9217-2a38854e19f0" />

Full Error

`Success
rows_processed: 0
rows_with_factors: 0
rows_without_factors: 0
rows_skipped: 50072
batches_processed: 0
[...]

PR #1225, opened 2026-05-19
fix(emission-recalc): savepoint-isolate per-entry recalc; abort batch on dead/poisoned session

guilbep:
## Context

Surfaced on **stage** after #1219 unblocked the recalc chain. `EmissionRecalculationWorkflow.recalculate_for_data_entry_type` caught every per-entry exception and continued **without rolling back `self.session`**. One DB failure poisoned the shared session, so every remaining entry re-raised *"transaction is aborted / can't reconnect"* — a log storm (observed: `data_entry_id 19903..19922`) and a job that failed while masking the first cause. The in-code comment even claimed a rollback that never happened.

## Changes

- **Per-entry `SAVEPOINT`** (`session.begin_nested()`): a bad row is contained and the loop's intended "skip & continue" behaviour now actually works (it was broken — the session stayed poisoned).
- **Fast-abort on a session/connection-fatal error**: re-raise so the runner (via #1219 Fix 1) records `FINISHED+ERROR` with the real cause instead of one identical fatal error per remaining entry. Fatal = `DBAPIError.connection_invalidated` (dropped connection) **or** `InvalidRequestError` incl. `PendingRollbackError` / "can't reconnect" (session needs a full rollback — even `begin_nested()`'s SAVEPOINT-enter fails on the next entry; this is the shape the stage log actually shows).
- Two regression tests (`connection_invalidated`, `PendingRollbackError`), **both verified to fail without the fatal clause**. Existing 10 workflow tests still green (12 total).

## Scope / honest limits

- This makes the recalc loop **resilient** to a transient DB fault (fail fast + clean, no storm, job correctly ERRORs).
- It does **not** explain or prevent the *first* `UndefinedTable: relation "data_ingestion_jobs" does not exist` seen simultaneously across three independent sessions on stage — that points to a concurrent DDL/deploy window (consistent with "table exists now / migration worked"). Tracked as a separate **environmental** question, not addressed here.
- The aggregation→carbon-report-stats path (`recompute_stats`/`update_stats`) has the **same unguarded-flush anti-pattern** (the earlier `InFailedSqlTransaction` on `carbon_report_modules`). Same SAVEPOINT treatment warranted — deliberately **not bundled**; follow-up PR.

## Verification

`cd backend && uv run pytest tests/unit/workflows/test_emission_recalculation.py` → 12 passed. Lint clean. Gold-standard: reverting the fatal clause makes both new regression tests fail.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/backend/app/api/v1/data_sync.py
+++ b/backend/app/api/v1/data_sync.py
@@ -2,7 +2,7 @@ import asyncio
 import json
 from datetime import datetime
 from typing import Optional
-from uuid import UUID
+from uuid import UUID, uuid4
 
 from fastapi import (
     APIRouter,
@@ -79,6 +79,7 @@ async def _stamp_job_type_and_meta(
     job_type: str,
     provider_name: Optional[str] = None,
     extra_meta: Optional[dict] = None,
+    pipeline_id: Optional[UUID] = None,
 ) -> None:
     """After ``provider.create_job`` returns, stamp ``job_type`` and
     extend ``meta`` so the runner's registry lookup hits the right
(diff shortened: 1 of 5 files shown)


guilbep:
not necessary anymore done mainly in  pr #1249

PR #1308, opened 2026-05-27
feat: Add columnSize support to module table

### LINKED ISSUE #1233: [FEAT](Design Refactor) Improve responsiveness of tables
Improve responsiveness of tables (Create a system through the module configuration where we could assign predefined sizes for each column with a fallback to default which would allow more flexibility and also avoid the cutting of words in dropdowns. That would be the right / longer way to do it) - as suggested by Ben in #690

BenBotros:
## What does this change?
Introduces a `columnSize` token system to replace ad-hoc `minColumnWidth`/`maxColumnWidth` magic numbers across module configs.

- `moduleConfig.ts`: adds a `COLUMN_SIZES` map (`xs: 80, sm: 120, md: 160, lg: 200, xl: 260`) and a `ColumnSize` type; adds `columnSize?: ColumnSize` to `ModuleField`
- `ModuleTable.vue`: updates `getColumnStyle` and `getColumnClasses` to resolve `columnSize` through `COLUMN_SIZES` as a fallback when `minColumnWidth` is not set; fixes truthy checks on width properties to use `!== undefined`
- Module configs (`buildings`, `equipment-electric-consumption`, `external-cloud-and-ai`, `headcount`, `process_emissions`, `purchase`): replaces hardcoded `minColumnWidth`/`maxColumnWidth` values with the appropriate `columnSize` token

## Why is this needed?
Column widths were scattered as raw pixel values across many module config files, making them hard to keep consistent and adjust globally. A shared token set makes sizing declarative and maintainable from one place.

## Type of change
- [x] 🔧 Configuration change
- [x] 🎨 Design/UI improvement
- [x] 🧹 Code cleanup

## Related issues
- Closes #1233

### DIFF
--- a/frontend/src/components/organisms/module/ModuleTable.vue
+++ b/frontend/src/components/organisms/module/ModuleTable.vue
@@ -71,8 +71,9 @@
           :align="col.align"
           class="q-pa-xs"
           :class="{
-            'column-max-width': col.maxColumnWidth,
-            'column-min-width': col.minColumnWidth,
+            'column-max-width': col.maxColumnWidth !== undefined,
+            'column-min-width':
+              col.minColumnWidth !== undefined || col.columnSize !== undefined,
           }"
           :style="getColumnStyle(col)"
         >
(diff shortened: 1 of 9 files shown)


guilbep (commented):
clean up old system

PR #1383, opened 2026-06-03
feat 1181: normalize equipment class/subclass and CSV lookup

### LINKED ISSUE #1181: [Bug]: (Equipment) - sorting classes alphabetically
### What went wrong?

 In the Equipment module, sorting classes alphabetically does not work

Image

### How can we reproduce this?

1. Go to Module Equipment
2. Add Scientific Equipment with the import functionnality (one euipement with Agitator/Incubator,  one Laminar flow hood, one Potentiostat)
3. Add manually some Scientific equipment with different classes in this sequence 2 3 times  with classe Agitator/Incubator, 12 times with Laminar flow hood, 2 times with Potentiostat, 1 time with Pneumatic, 2 times with audio)
4. Sort in the table by classes

### Screenshot

_No response_

### Which browser are you using?

Chrome

### What device are you using?

Desktop/Laptop

### Additional information

_No response_

### Before submitting

- [x] I've checked that this issue hasn't already been reported

BenBotros:
## What does this change?
Normalizes equipment class/subclass values during CSV import so that human-readable labels in uploaded files match internal factor keys, and eliminates split sort groups in the equipment table.

- `seed_helper.py`: adds `normalize_kind_for_lookup()` which strips internal prefixes (`equipment_factor_`, `equipment_factor_sub_`) and removes all non-alphanumeric characters; adds a third lookup strategy (Strategy 3) in `load_factors_map` that registers each factor under a normalized key (`_n_:<type>:<year>:<norm_kind>:<norm_subkind>`) in addition to the existing exact-match keys
- `base_csv_provider.py`: adds a fallback factor lookup using `normalize_kind_for_lookup` when the exact-match lookup fails (e.g. CSV row has `"Agitator / Incubator"` but the factor key is `"equipment_factor_agitator_incubator"`); once a factor is resolved, rewrites the `kind_field` and `subkind_field` values in the row to the canonical values from the factor's classification, so imported entries always use the same key as manually-created entries
- `schemas.py` (equipment module): updates the `select_map` to normalize `equipment_class` and `sub_class` from `DataEntry.data` at query time using `func.lower` + `func.regexp_replace` (stripping the `equipment_factor_` prefix and non-alphanumeric chars), so that entries created before this fix also sort and group consistently; updates `filter_map` to read from `DataEntry.data` instead of `Factor.classification`

## Why is this needed?
Equipment entries imported via CSV used human-readable labels (`"Agitator / Incubator"`) while manually-created entries used internal factor keys (`"equipment_factor_agitator_incubator"`). This caused them to appear as separate sort groups in the table, and CSV imports silently dropped the factor association when the label didn't match exactly.

## Type of change
- [x] 🐛 Bug fix

## Code quality checklist

## Testing checklist

## Related issues
- Closes #1181

### DIFF
--- a/backend/app/modules/equipment_electric_consumption/schemas.py
+++ b/backend/app/modules/equipment_electric_consumption/schemas.py
@@ -1,6 +1,7 @@
 from typing import Optional, Self
 
 from pydantic import ValidationInfo, field_validator, model_validator
+from sqlalchemy import func
 
 from app.core.config import get_settings
 from app.core.logging import get_logger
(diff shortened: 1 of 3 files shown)


guilbep:
not GONNA MERGE:

done already in * c8f95d3c fix(1181): use data_entry for sort

PR #1384, opened 2026-06-03
feat 863: add class to plane trip emission calculation

### LINKED ISSUE #863: [FEAT](Travel) Add the class in the air travel emissions calculations
The factors we gave you previously did not distinguish between economy/business/first class in air travel. We have now factor that do the differentiation and we'd like to add this difference to the formula that calculates the air travel emissions for flights inserted manually in the calculator. 

@charlottegiseleweil

BenBotros:
## What does this change?
Adds cabin class to plane trip emission calculations and updates the haul category bucketing to match the new two-bucket factor structure.

- `distance_geography.py`: adds `get_emission_haul_category()` with a two-bucket split at 2500 km (`short_to_medium_haul` / `medium_to_long_haul`), replacing the old four-bucket `get_haul_category()` for emission factor lookups
- `schemas.py` (professional travel): sets `subkind_field = "class"` on the plane handler so cabin class is used as the factor subkind; maps the UI value `"eco"` to the factor CSV value `"economy"` in `resolve_computations`; fixes the multiplier key from `rfi_adjustment` to `rfi_adjustement` (matching the CSV); updates `TravelPlaneBase` to declare `class_` with `Field(alias="class")` (Python reserved word workaround); updates valid categories to the new two-bucket values; adds `validate_class` validator; removes the stale `class_adjustement` field; adds `ConfigDict(populate_by_name=True)` to the three factor DTOs
- `factor.py` (schema): updates `expected_columns` and `required_columns` to use the field's alias when defined, so `"class"` is correctly identified as an expected column
- `classification_fields` for `TravelPlaneFactorHandler`: adds `"class"` alongside `"category"`
- Tests: updates all plane factor fixtures from `very_short_haul` to `short_to_medium_haul` + `class: economy`; adds `TestGetEmissionHaulCategory` unit tests; fixes multiplier key and factor values in integration test fixtures

## Why is this needed?
Plane emission factors are differentiated by both distance category and cabin class, but the emission calculation was ignoring cabin class entirely. The factor structure has also changed from four haul buckets to two, requiring the distance-to-category mapping to be updated.

## Type of change
- [x] ✨ New feature
- [x] 🐛 Bug fix

## Code quality checklist

## Testing checklist

## Related issues
- Closes #863

### DIFF
--- a/docs/src/backend/06-PERMISSION-SYSTEM.md
+++ b/docs/src/backend/06-PERMISSION-SYSTEM.md
@@ -164,7 +164,7 @@ canonical identifiers live in the `RoleName` enum in
 - `calco2.user.principal` — Unit manager, unit-scope access (all modules).
 - `calco2.backoffice.metier` — Backoffice administrator: affiliation-scoped
   reporting plus scope-less users / documentation / ui_texts.
-- `calco2.superadmin` — Super administrator with every backoffice page,
+- `calco2.backoffice.admin` — Super administrator with every backoffice page,
   including configuration / pipeline_operations / logs (no `modules.*` grants).
 
 ### Scopes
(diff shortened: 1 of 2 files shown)


backend/app/modules/professional_travel/schemas.py
```diff
@@ -249,6 +249,7 @@ class ProfessionalTravelPlaneModuleHandler(ProfessionalTravelBaseModuleHandler):
     response_dto = ProfessionalTravelPlaneHandlerResponse
 
     kind_field: str = "category"
+    subkind_field: str = "class"
```
guilbep: cabin_class cf doc! @BenBotros

backend/app/utils/distance_geography.py
```diff
@@ -128,6 +128,11 @@ def get_haul_category(distance_km: float) -> str:
         return "long_haul"
 
 
+def get_emission_haul_category(distance_km: float) -> str:
+    """Return the factor category for emission lookup, split at 2500 km."""
+    return "short_to_medium_haul" if distance_km < 2500 else "medium_to_long_haul"
```
guilbep: @BenBotros  ???
ok.. I'm pretty sure we should use the factors for that...

guilbep (changes requested):

guilbep:
Use distances in the factors.csv instead of hardcoding

PR #1537, opened 2026-06-16
feat 1530: derive IN_PROGRESS status for modules with data

### LINKED ISSUE #1530: [BUG](Navigation bar) Status icons are missing upon first navigation
Upon first navigation on a module where data was loaded from the backoffice, we should see status `In progress`. Indeed status is visible as "in progress" in the top left , but the orange icon is missing until we validate it for the first time. 

Image

BenBotros:
## What does this change?
Derives an effective `IN_PROGRESS` status for navigation/listing purposes when a module already holds data but has never been validated.

- `carbon_report_module_service.py`: `list_modules` now counts data entries per module (via new `_entry_counts_by_module` batch query) and upgrades any `NOT_STARTED` module with at least one entry to `IN_PROGRESS` in the returned read models. The stored DB status is left untouched — the report's `overall_status` rollup keys off `VALIDATED` counts only, so it is unaffected by this change
- Adds two unit tests: `test_list_modules_derives_in_progress_from_data_entries` (a module with an unvalidated data entry surfaces as `IN_PROGRESS`; untouched modules stay `NOT_STARTED`) and `test_list_modules_keeps_validated_status_over_data_derivation` (a `VALIDATED` module with data is not downgraded)

## Why is this needed?
Modules populated with data (e.g. imported from the backoffice) but never explicitly validated remained `NOT_STARTED` in the database, since nothing bumps the status on write. This left the navigation bar's status icon missing even though the module was, in fact, in progress.

## Type of change
- [x] 🐛 Bug fix

## Code quality checklist

## Testing checklist

## Related issues
- Closes #1530

### DIFF
--- a/backend/app/repositories/carbon_report_module_repo.py
+++ b/backend/app/repositories/carbon_report_module_repo.py
@@ -195,6 +195,45 @@ class CarbonReportModuleRepository:
         await self.session.refresh(db_obj)
         return db_obj
 
+    async def get_report_ids_with_entries_by_status(
+        self,
+        module_type_id: int,
+        year: int,
+        status: int,
+        carbon_report_module_id: Optional[int] = None,
+    ) -> List[int]:
+        """Distinct carbon_report_ids whose ``module_type_id`` module in
+        ``year`` has the given ``status`` and at least one data entry.
+
+        Used by the ingestion tasks to bump freshly-populated NOT_STARTED
+        modules to IN_PROGRESS (#1530) without embedding SQL in the task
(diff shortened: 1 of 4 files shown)


guilbep (changes requested):
don't do it like that @BenBotros 

update by using update_status() in carbon_report_module_repo in ingestion_tasks when  uploading data.csv and/or sync api travel

guilbep:
ALREADY MERGED. WONT MERGED

PR #1626, opened 2026-06-22
feat(i18n): add new refrigerant category translations for emissions module

### LINKED ISSUE #1446: [Bug]: (Process Emission) - Import file with Refrigerant/Fluide Frigorigène
### What went wrong?

Problème d'import de fichier avec le gas/gaz "Refrigerant/Fluide Frigorigène"
Dans le csv : 
-si la catégorie est en EN (Refrigerant), cela affiche ensuite cette colonne fausse process-emissions.category.refrigerant dans le tableau. 
-Si la catégorie est en FR (Fluide frigorigène), un message d'erreur apparait

### How can we reproduce this?

1.Créer un fichier ep5 avec catégorie Refrigerant et dans le champ sous-catégorie le texte "R11", et quantité 51. 
2.Go to module Process Emission, puis dans Ajouter un gaz et  Cliquer sur Importer CSV

### Screenshot

Image

### Which browser are you using?

Chrome

### What device are you using?

Desktop/Laptop

### Additional information

@Anna-Kounina-Masse for information

### Before submitting

- [x] I've checked that this issue hasn't already been reported

BenBotros:
## What does this change?
Adds a capitalized-key variant of the Refrigerant i18n entry in `process_emissions.ts`.

- Adds `process-emissions.category.Refrigerant` (capital R) alongside the existing `process-emissions.category.refrigerants` and `process-emissions.category.refrigerant` (lowercase) keys, all mapping to the same EN/FR label

## Why is this needed?
The CSV import uses `"Refrigerant"` (capital R) as the category value, but the i18n lookup was only registered for the lowercase variants. The missing key caused the label to fall back to the raw key string in the UI after import.

## Type of change
- [x] 🐛 Bug fix

### DIFF
--- a/frontend/src/i18n/process_emissions.ts
+++ b/frontend/src/i18n/process_emissions.ts
@@ -57,6 +57,10 @@ export default {
     en: 'Refrigerant',
     fr: 'Fluide Frigorigène',
   },
+  [`${MODULES.ProcessEmissions}.category.Refrigerant`]: {
+    en: 'Refrigerant',
+    fr: 'Fluide Frigorigène',
+  },
   // Factor taxonomy / CSV use singular "refrigerant"; same label as plural key above.
   [`${MODULES.ProcessEmissions}.category.refrigerant`]: {
     en: 'Refrigerant',


frontend/src/i18n/process_emissions.ts
```diff
@@ -57,6 +57,10 @@ export default {
     en: 'Refrigerant',
     fr: 'Fluide Frigorigène',
   },
+  [`${MODULES.ProcessEmissions}.category.Refrigerant`]: {
```
ymarcon, line 60: Refrigerant or refrigerant ?

ymarcon (changes requested):

frontend/src/i18n/process_emissions.ts
BenBotros: It's correct I purposely made a fallback because they added 'Refrigerant' in data/templates/factors

ymarcon (approved):

PR #1686, opened 2026-06-29
feat/1416 feat navigation design refactor

### LINKED ISSUE #1416: [FEAT] Navigation Design Refactor
# Success criteria

- [x] Implementation of the new workflow according to this design
- [x] Text in "Ready to map" should be different for `User standard` and `User principal`:
- `User standard` : "You are user standard, so you can see only a few modules to fit data privacy constraints. See (i) About button for details, and request full access."
- `User principal` :  "You are user principal, so you can see all modules . If relevant, the Head of Unit can delegate this role to team members. See (i) About button for details."
- Text for (i) About : "Processus to give access in ACCRED blablabla- Mention careful to data privacy!"

Image
Image

# Goal

Collapse the three separate screens (`workspace-setup`, calculator `home`, simulator `home`) into **one page** so the user lands directly in their workspace without a setup step. Keep all existing content only the layout and entry flow change.

---

### 1. Single merged route

- Replace the three routes with one page (e.g. `/:lang/:unit/:year`).
- Order top → bottom: **header context bar → Calculator (primary) → Simulator (secondary)**.
- Calculator gets the visual priority (full width, leading position); Simulator sits below in a visually de-emphasised section (tinted band, smaller heading).

### 2. Move unit / year selection into the header

- Remove the standalone "Workspace Setup" screen.
- Add two `<select>` controls in a sticky header control bar: **Unit / Lab** and **Assessment year** (year option labels show the total, e.g. `2026 — 26 t CO₂-eq`).
- Changing either select updates the page in place (re-fetch the unit/year data, update URL params) — no page navigation, no separate "Change" button.
- Show the current user identity in the header: manager name + `Unit Manager` role pill + affiliation breadcrumb (`EPFL › ENAC-TEST › … › ENAC-IT4R-TEST`, current unit emphasised). No avatar icon.
- Module-completion progress ("This year 4/8", "Last year 0/8") moves **out of the header** into the Calculator section header.

### 3. Hide the long welcome copy behind a popup

- The large "Welcome to the CO₂ Calculator" methodology/instructions text should **not** sit inline taking vertical space.
- Replace with a short intro + a "Learn more" / info trigger that opens the full text in a **modal/popover**.
- Same treatment for the Simulator intro paragraph.

BenBotros:
## What does this change?
Major navigation redesign: replaces the multi-step `WorkspaceSetupPage` with a unified home page that hosts inline unit/year dropdowns, an emission breakdown chart with clickable module icon axes, a calculator updates panel, a CO₂ Project Planner section, and a CO₂ Explorer section.

**Router refactor:**
- `workspace-setup` route and `WorkspaceSetupPage.vue` deleted; replaced by a parameterless landing route (`home-landing`) whose sole job is to run `redirectToWorkspaceIfSelectedGuard` and forward to the unified home page
- `WORKSPACE_SETUP_ROUTE_NAME` → `HOME_LANDING_ROUTE_NAME`; `DEFAULT_ROUTE_NAME` now points to the new landing route
- `SimulationsPage.vue` deleted; the `simulation` route now redirects to the unified home page; a new `simulation-add` route is added pointing to a new `AddSimulationPage.vue`
- All guards and error pages updated to use `DEFAULT_ROUTE_NAME` instead of `workspace-setup`
- `validateUnitGuard` refactored: the workspace-loading logic extracted to a shared `loadWorkspaceFromRoute()` function used by both `beforeEnter` and `WorkspacePage.vue`'s new `onBeforeRouteUpdate` (so param-only navigation via the home dropdowns re-validates the workspace without re-running `beforeEnter`)
- `redirectToWorkspaceIfSelectedGuard` extended: instead of just forwarding to `home` when a selection is persisted, it now also resolves a default workspace (first unit + most recent open year) when no selection exists, and redirects to `/unauthorized` when the account has no units

**New/modified components:**
- `WorkspaceSelectorBar.vue` (new): inline unit and year dropdowns with affiliation breadcrumb and principal user display; fetches carbon reports on mount and on unit change
- `CO2ProjectPlanner.vue` (new): full-width grey-band section with a project list table (static placeholder rows) and "Start a project" button
- `CO2Explorer.vue` (new): section with icon, title, description, and "Start an Exploration" button
- `ModuleIconBox.vue`: adds `submoduleType` prop passed through to `getModuleIconColors` for buildings disambiguation
- `ModuleCarbonFootprintChart.vue`: adds `hideHeader`, `hideActions`, and `moduleIconAxis` props; when `moduleIconAxis` is true, text x-axis labels are hidden and replaced by an absolutely-positioned overlay of clickable `ModuleIconBox` buttons aligned under each bar via `convertToPixel`; bar width fixed to 40px in icon-axis mode; `CATEGORY_TO_SUBMODULE` map handles the buildings two-bar case
- `Co2ModuleSidebar.vue`: adds a "Home" item at the top of the nav list (same grey pill style as the documentation button, with collapsed tooltip)
- `Co2Header.vue`: `logoRoute` now returns the home route when unit+year are in the URL; "Change workspace" button links to home (not workspace-setup); breadcrumb home link updated; workspace bar suppressed on the `home` route instead of `workspace-setup`; imports `DEFAULT_ROUTE_NAME`
- `Co2Header.stories.ts`: deleted
- `LabSelectorItem.vue`: i18n keys renamed `workspace_setup_unit_affiliation` → `workspace_unit_affiliation`, `workspace_setup_unit_progress` → `workspace_unit_progress`
- `YearSelector.vue`: i18n key renamed `workspace_setup_year_table_header_year` → `workspace_year_table_header_year`
- `ErrorNotFound.vue` / `ErrorUnauthorized.vue`: use `DEFAULT_ROUTE_NAME` instead of `WORKSPACE_SETUP_ROUTE_NAME`

**`HomePage.vue` rewrite:**
- Removed: module cards grid, status badges, `home_intro_2`/`home_intro_3`/`home_intro_4` paragraphs that described the old multi-step flow, `formatTonnesCO2` import
- Added: sticky `WorkspaceSelectorBar` pinned under the header; a two-column `calculator-card` (left: title/chart/footer, right: updates); role badge with access popover (standard vs principal, ACCRED link); `ModuleCarbonFootprintChart` in `moduleIconAxis` mode when validated data exists, otherwise an empty-state CTA; `CO2ProjectPlanner` (full-width band); `CO2Explorer` section; lazy-loaded chart chunk via `defineAsyncComponent`; `fetchEmissionBreakdown` called on mount and on carbon report change

**i18n:**
- `workspace_setup.ts` deleted; all keys migrated to `home.ts` under `workspace_*` prefix (dropping `_setup_` segment)
- `home.ts`: large addition of unified home page keys (`co2_calculator_*`, `co2_project_planner_*`, `co2_explorer_*`, `planner_table_*`, `workspace_*` workspace-selector keys, `home_btn`)
- `calculator_update.ts`: adds three placeholder update entries (`calculator_update_1/2/3_title/date/body`)
- `common.ts`: adds `home_btn`

## Why is this needed?
The multi-step workspace-setup flow required users to select unit, workspace type, and year before reaching the calculator — adding friction on every visit. The unified home page collapses this into inline dropdowns always visible in a sticky bar, shows the emission chart and updates at a glance, and surfaces the simulator/planner tools without a separate landing step.

## Type of change
- [x] ✨ New feature
- [x] 🎨 Design/UI improvement
- [x] 🧹 Code cleanup


frontend/src/router/routes.ts
```diff
@@ -22,7 +22,7 @@ export {
   LOGIN_TEST_ROUTE_NAME,
   LOGIN_ROUTES,
   HOME_ROUTE_NAME,
-  WORKSPACE_SETUP_ROUTE_NAME,
+  HOME_LANDING_ROUTE_NAME,
```
guilbep: @BenBotros  HOME_LANDING_ROUTE_NAME est import mais n'est pas utilisé ?

frontend/src/router/routes.ts
```diff
@@ -33,7 +33,6 @@ import {
   LOGIN_ROUTE_NAME,
   LOGIN_TEST_ROUTE_NAME,
   HOME_ROUTE_NAME,
-  WORKSPACE_SETUP_ROUTE_NAME,
```
guilbep, line 36: * **Unify routeNames usage and remove duplicate module edges**

  ```ts
  import * as RouteNames from './routeNames';
  ```

  Remove:

  ```ts
  export { ... } from './routeNames';
  import { ... } from './routeNames';
  ```

  → single dependency path, no redundant imports/exports in `routes.ts`.

---

* **Make i18n fully URL-driven with browser fallback only at entry/guard level**

  ```ts
  router.beforeEach((to) => {
    const urlLang = to.params.language;

    const lang =
      typeof urlLang === 'string'
        ? urlLang
        : getBrowserLocale();

    i18n.global.locale.value = lang;
  });
  ```

  Root redirect uses browser only when URL is missing:

  ```ts
  redirect: () => ({
    name: RouteNames.DEFAULT_ROUTE_NAME,
    params: { language: getBrowserLocale() },
  })
  ```

  → URL = source of truth, browser = initial seed only.

---

* **Enforce strict separation: routes = declarative, guards = runtime logic**

  ```ts
  // routes.ts (pure structure only)
  {
    path: '/:language',
    children: [...]
  }
  ```

  ```ts
  // router/guards.ts (behavior layer)
  router.beforeEach(i18nSyncGuard);
  router.beforeEach(authGuard);
  router.beforeEach(permissionGuard);
  ```

  → no state access in route definitions, all derivation happens at navigation time.

frontend/src/router/routes.ts
```diff
@@ -206,12 +208,25 @@ const routes: RouteRecordRaw[] = [
...
+                redirect: (to) => ({
+                  name: HOME_ROUTE_NAME,
+                  params: to.params,
+                }),
+              },
+              {
+                // Reached from the "Start a project" button on the unified
+                // home page (CO2ProjectPlanner).
+                path: 'simulation/add',
+                name: 'simulation-add',
```
guilbep: @BenBotros  if something is called planner, maybe rename the path/anme component to reflect the name ?

frontend/src/router/guards/validateUnitGuard.ts
```diff
@@ -1,5 +1,7 @@
+import type { RouteLocationNormalized } from 'vue-router';
 import { useTimelineStore } from 'src/stores/modules';
 import { useWorkspaceStore } from 'src/stores/workspace';
+import { DEFAULT_ROUTE_NAME } from '../routeNames';
```
guilbep: @BenBotros  pas de relatif '../'

frontend/src/router/guards/validateUnitGuard.ts
```diff
@@ -50,7 +62,7 @@ export default async function validateUnitGuard(to) {
   // then we can retrieve modules
   if (!response && !carbonReportId) {
     return {
-      name: 'workspace-setup',
+      name: DEFAULT_ROUTE_NAME,
       params: {
         language: to.params.language || 'en',
```
guilbep: same here, I'm pretty sure I was a bit lazy when I did that.

we should check every guard and prams.language in the code base to avoid having 'en' hard-coded


* **Stop hard-coding fallback language in navigation params**

  ```ts
  params: {
    language: to.params.language || 'en',
  }
  ```

  → this bakes policy into every redirect and scatters defaulting logic.

---

* **Centralize locale resolution (URL → fallback chain)**

  ```ts
  const resolveLanguage = (to: RouteLocationNormalized) => {
    const url = to.params.language;
    if (typeof url === 'string' && url.length) return url;

    return getBrowserLocale();
  };
  ```

  usage:

  ```ts
  params: {
    language: resolveLanguage(to),
  }
  ```

  → single rule for URL parsing + fallback (no inline `'en'` anywhere).

---

* **Keep redirects stateless + deterministic**

  ```ts
  redirect: (to) => ({
    name: RouteNames.DEFAULT_ROUTE_NAME,
    params: {
      language: resolveLanguage(to),
    },
  })
  ```

  → avoids duplicated fallback logic across routes, keeps routing behavior consistent with the same resolution layer used in `beforeEach`.

frontend/src/router/guards/authGuard.ts
```diff
@@ -28,7 +28,7 @@ export async function authGuard(to: RouteLocationNormalized) {
 
   // Redirect authenticated users away from login
   if (LOGIN_ROUTES.includes(to.name as string) && auth.isAuthenticated) {
-    return { name: 'workspace-setup', ...redirectTo };
+    return { name: DEFAULT_ROUTE_NAME, ...redirectTo };
```
guilbep, line 31: 👍

frontend/src/router/guards/redirectToWorkspaceIfSelectedGuard.ts
```diff
@@ -1,14 +1,35 @@
...
+/**
+ * Resolver for the parameterless landing route. The unified home page hosts the
+ * Unit/Year dropdowns, so this guard's only job is to pick a default workspace
+ * and forward to the home page:
+ *
+ *   1. A persisted selection (returning user) wins.
+ *   2. Otherwise the first unit + most recent open year is chosen.
+ *
+ * When the account has no units there is nothing to show, so we forward to
+ * /unauthorized. `validateUnitGuard` redirects here with `?unit&year` (both
```
guilbep: ?? pas possible

guilbep: rajouter des tests pour les cas suivants:


1) no units
2) severals units
3) no year
5) multiple year
6) one year

frontend/src/router/routes.ts
```diff
@@ -131,8 +130,22 @@ const routes: RouteRecordRaw[] = [
...
-        redirect: { name: DEFAULT_ROUTE_NAME },
         children: [
+          {
+            // Parameterless landing (default route). It never renders: its
+            // `beforeEnter` resolves a default unit/year and forwards to the
+            // unified home page (or to /unauthorized when the account has no
+            // units). Replaces the former interactive `workspace-setup` page.
+            path: '',
+            name: DEFAULT_ROUTE_NAME,
+            beforeEnter: redirectToWorkspaceIfSelectedGuard,
```
guilbep: rename to redirecToDefaultRoute 

be consistent

remove if you think necessary

frontend/src/pages/app/WorkspacePage.vue
```diff
@@ -12,6 +15,22 @@ const fetchYearConfig = async () => {
 onMounted(async () => {
   await fetchYearConfig();
 });
+
+// `beforeEnter` does not re-run on param-only navigation, so switching unit or
+// year via the home-page dropdowns is handled here: re-validate the workspace
+// and reload the year configuration before the updated route resolves.
+onBeforeRouteUpdate(async (to, from) => {
```
guilbep: @BenBotros thinks it's not necessary

frontend/src/router/guards/redirectToWorkspaceIfSelectedGuard.ts
```diff
@@ -1,14 +1,35 @@
...
+ * null) when the persisted unit is no longer valid — that path resets the
+ * selection first so we resolve a fresh default instead of looping back to the
+ * stale unit.
+ */
+function unitSlug(id: number, name: string): string {
+  return `${id}-${name.replace(/\s+/g, '-').toLowerCase()}`;
```
guilbep: not dry: already exist in router/guards
make it dry

frontend/src/router/guards/redirectToDefaultRoute.ts
```diff
@@ -19,5 +40,33 @@ export default async function redirectToWorkspaceIfSelectedGuard(to) {
...
+
+  const started = yearConfigStore.startedYears;
+  const reportYears = workspaceStore.carbonReports.map((report) => report.year);
+  const openYears = reportYears.filter((year) => started.has(year));
+  const year =
+    openYears.length > 0
+      ? Math.max(...openYears)
+      : reportYears.length > 0
+        ? Math.max(...reportYears)
+        : new Date().getFullYear() - 1;
```
guilbep: rename file, and abstract in small small function

one function does one thing ==> better naming, better comprehension


Good. This is exactly the case where you want **pure single-responsibility functions + a strict failure policy**.

Here’s a clean decomposition where each function does *one thing only* and nothing guesses anything.

---

## 1. Unit resolution (must exist)

```ts
async function resolvePrimaryUnit(workspaceStore: any) {
  await workspaceStore.getUnits();

  const unit = workspaceStore.units[0];
  if (!unit) {
    throw new Error('No units available: unauthorized');
  }

  return unit;
}
```

---

## 2. Data loading (no logic, just fetch)

```ts
async function loadYearData(
  workspaceStore: any,
  yearConfigStore: any,
  unitId: string,
) {
  await Promise.all([
    workspaceStore.fetchCarbonReportsForUnit(unitId),
    yearConfigStore.fetchConfiguredYears(),
  ]);
}
```

---

## 3. Year extraction (pure logic, strict)

```ts
function extractOpenYears(
  reportYears: number[],
  startedYears: Set<number>,
) {
  const openYears = reportYears.filter(y => startedYears.has(y));

  if (openYears.length === 0) {
    throw new Error('No open years available');
  }

  return openYears;
}
```

---

## 4. Final selection (pure deterministic logic)

```ts
function pickLatestYear(years: number[]) {
  if (years.length === 0) {
    throw new Error('No years available');
  }

  return Math.max(...years);
}
```

---

## 5. Orchestrator (clean, readable, no logic inside)

```ts
async function resolveYearContext(workspaceStore: any, yearConfigStore: any) {
  const unit = await resolvePrimaryUnit(workspaceStore);

  await loadYearData(workspaceStore, yearConfigStore, unit.id);

  const reportYears = workspaceStore.carbonReports.map(r => r.year);
  const started = yearConfigStore.startedYears;

  const openYears = extractOpenYears(reportYears, started);

  const year = pickLatestYear(openYears);

  return { unit, year };
}
```

---

## Why this is better

* each function has exactly **one responsibility**
* no hidden fallback logic
* no branching in the orchestrator
* errors happen **at the right semantic level**
* logic is now testable in isolation

---

If you want to go one step further, I can make this **fully functional (no stores inside helpers)** or convert it into a **composable Vue hook (`useResolvedYearContext`)**.

frontend/src/pages/app/HomePage.vue
```diff
@@ -46,289 +73,496 @@ const firstEditableModule = computed(() =>
...
+const accessBodyKey = computed(() =>
+  isPrincipalUser.value
+    ? 'co2_calculator_access_principal_body'
+    : 'co2_calculator_access_standard_body',
 );
 
-const modulesCounterText = computed(() => t('home_modules_counter'));
+// EPFL access-management portal, opened straight on the authorizations tab.
+// Principals delegate roles here; standard users instead email the principal.
+const ACCRED_URL = 'https://accred.epfl.ch?opentab=authorizations';
```
guilbep: 😿 

rename to something more generic like AUTHORIZATION_PROVIDER_URL and use .env.example and env variable of the frontend (cf Dockerfile)

frontend/src/components/charts/results/ModuleCarbonFootprintChart.vue
```diff
@@ -104,6 +125,53 @@ const labelToKey = computed<Record<string, string>>(() => {
...
+    if (iconAxisItems.value.length) iconAxisItems.value = [];
+    return;
+  }
+  const items = datasetSource.value;
+  if (!items.length) {
+    iconAxisItems.value = [];
+    return;
+  }
+
+  const yZero = chart.convertToPixel({ yAxisIndex: 0 }, 0) as number;
```
guilbep, line 185: @BenBotros  create a new issue refactor, so we can talk on how not to do that? if possible please

frontend/src/components/charts/results/ModuleCarbonFootprintChart.vue
```diff
@@ -1443,16 +1520,45 @@ const downloadCSV = () => {
...
+            v-for="item in iconAxisItems"
+            :key="item.label"
+            class="module-icon-axis__item"
+            :class="{ 'module-icon-axis__item--link': item.module }"
+            :to="
+              item.module
+                ? { name: 'module', params: { module: item.module } }
+                : undefined
+            "
+            :style="{ left: `${item.x}px`, top: `${iconAxisTop + 22}px` }"
```
guilbep: @BenBotros  maybe use class and custom properties instead of hard coded style ? 
not sure if relevant though. Maybe not even possible ? 
discard if not possible

frontend/src/components/charts/results/ModuleCarbonFootprintChart.vue
```diff
@@ -1443,16 +1520,45 @@ const downloadCSV = () => {
...
+          >
+            <ModuleIconBox
+              :name="item.module ?? 'addition-datas'"
```
guilbep: ? no magic constant? try putting that into named constant at the top of the file ?

frontend/src/components/layout/Co2Header.vue
```diff
@@ -62,23 +62,24 @@ const breadcrumbLabel = computed(() =>
...
-  const routeName = String(route.name ?? '');
-  const isSimulation =
-    routeName === 'simulation' ||
-    routeName === 'simulation-explore' ||
-    routeName === 'simulation-plan';
-
-  if (isSimulation) {
+  // Inside a workspace (unit + year present) the logo returns to the unified
+  // home page; otherwise it falls back to the landing resolver.
+  if (route.params.unit && route.params.year) {
```
guilbep, line 71: @BenBotros  maybe possible to use route meta to infer the default home page?
that way only one route?

super weird to have a if there ? seems like bad code,

should always be name: DEFAULT_ROUTE_NAME with 

params: {...params} ?

not sure I understand

frontend/src/components/layout/Co2Header.vue
```diff
@@ -128,14 +129,17 @@ const logoRoute = computed(() => {
...
             language: route.params.language || 'en',
           },
         }"
       />
 
-      <template v-if="route.name !== 'workspace-setup'">
+      <!-- The unified home page hosts the Unit/Year dropdowns itself, so the
+           workspace summary + "Change workspace" shortcut are suppressed there
+           and only shown on the module / results / simulation routes. -->
+      <template v-if="route.name !== 'home'">
```
guilbep: use constant from routeNames.ts

frontend/src/components/organisms/home/CO2ProjectPlanner.vue
```diff
@@ -0,0 +1,218 @@
...
+    align: 'right',
+  },
+]);
+
+// PLACEHOLDER: TO BE DELETED WHEN PLAN ROWS ARE POPULATED
+const planRows = [
+  {
+    name: 'Scientific Grant H',
+    date: '21-03-2026',
+    creator: 'Charlie Weil',
```
guilbep, line 50: @BenBotros are we sure we want to use our real names?

frontend/src/components/organisms/layout/Co2ModuleSidebar.vue
```diff
@@ -89,6 +92,10 @@ function navigateToModule(moduleLink: string) {
 function navigateToResults() {
   router.push({ name: 'results', params: route.params });
 }
+
+function navigateToHome() {
+  router.push({ name: HOME_ROUTE_NAME, params: route.params });
```
guilbep, line 97: @BenBotros  default_ROUTE?

frontend/src/pages/app/HomePage.vue
```diff
@@ -46,289 +73,496 @@ const firstEditableModule = computed(() =>
...
 
-const moduleCardsWithStatus = computed(() =>
-  MODULE_CARDS.filter(
-    (card) =>
-      yearConfigStore.isModuleVisible(card.module) &&
-      authStore.canUserAccessModule(card.module),
-  ),
+const accessTitleKey = computed(() =>
+  isPrincipalUser.value
+    ? 'co2_calculator_access_principal_title'
```
guilbep: @BenBotros  maybe possible to directly use without computed in the template using 

{{ $t(`co2_calculator_access_${user_type}_title`) }}

et tu fais juste une computed qui retourne standard/principale ?

frontend/src/pages/app/HomePage.vue
```diff
@@ -46,289 +73,496 @@ const firstEditableModule = computed(() =>
...
+});
 
-function getStatusBadgeVariant(moduleLink: Module): StatusBadgeVariant | null {
-  const state = timelineStore.itemStates[moduleLink];
-  if (state === MODULE_STATES.Validated) return 'validated';
-  if (state === MODULE_STATES.InProgress) return 'progress';
-  return null;
-}
+const calculatorUpdates = computed(() =>
+  [1, 2, 3]
```
guilbep: @BenBotros  j'aime pas trop les trucs hard-codé comme ça.

faut trouver une meilleure solution

un truc dynamique qui récupère toutes les clefs i18n ?

frontend/src/pages/app/HomePage.vue
```diff
@@ -46,289 +73,496 @@ const firstEditableModule = computed(() =>
...
+async function fetchEmissionBreakdown() {
+  const carbonReportId = workspaceStore.selectedCarbonReport?.id;
+  if (!carbonReportId) return;
+  await moduleStore.getEmissionBreakdown(carbonReportId, []);
 }
+
+onMounted(fetchEmissionBreakdown);
+
+watch(
+  () => workspaceStore.selectedCarbonReport?.id,
```
guilbep: @BenBotros est-ce qu'on peut remplacer ça par un autre event de routeUpdate? plutôt qu'un watch ?

guilbep (changes requested):

frontend/src/components/organisms/home/CO2ProjectPlanner.vue
BenBotros: Yes we wanted to identify who created the plan

helm/values.yaml
```diff
@@ -225,6 +225,12 @@ frontend:
     # Environment label attached to every Sentry event. Override per-cluster
     # in the ops repo (e.g. "stage", "production").
     APP_ENVIRONMENT: "production"
+    # Access-management provider shown in the calculator's access popover
+    # (role delegation). Defaults in src/config/runtime.ts point at EPFL's
+    # ACCRED; override here (or in the ops repo) to rebrand for another
+    # institution.
+    APP_AUTHORIZATION_PROVIDER_NAME: "ACCRED"
+    APP_AUTHORIZATION_PROVIDER_URL: "https://accred.epfl.ch?opentab=authorizations"
```
guilbep: change env variable name to be more generic

APP_ACCESS_MANAGEMENT_PROVIDER_NAME
= public frontend label

APP_ACCESS_MANAGEMENT_PROVIDER_URL
= public frontend href/button URL

don't put REAL url inside the values.yaml it's an example.

we need to set the real env variable in dev/stage/prod git-ops repo

btw, right now it's accred but maybe it's something else? like the role_provider in the backend ?

(I guess jwt (default)/accred/test))

maybe add a comment explaining what other values instead of 'accred' is possible.
Also, weird, it's 'accred'. in backend and 'ACCRED' in frontend?

frontend/.env.example
```diff
@@ -13,3 +13,9 @@ APP_SENTRY_DSN=
 # Environment label attached to every event. Conventional values: development,
 # stage, production. Lets you filter events per env in GlitchTip.
 APP_ENVIRONMENT=development
+
+# Access-management provider shown in the calculator's access popover (role
+# delegation). Empty / unset → defaults in src/config/runtime.ts (EPFL's
+# ACCRED). Override to rebrand for another institution.
+APP_AUTHORIZATION_PROVIDER_NAME=ACCRED
+APP_AUTHORIZATION_PROVIDER_URL=https://accred.epfl.ch?opentab=authorizations
```
guilbep: @BenBotros don't put real values inside btw!

guilbep (changes requested):

frontend/src/router/guards/redirectToDefaultRoute.ts
```diff
@@ -0,0 +1,128 @@
...
+ * Pick the year to land on, preferring the most recent year that is both
+ * reported and still open for editing, then the most recent reported year, and
+ * finally last calendar year when the unit has no reports at all.
+ */
+function pickDefaultYear(
+  reportYears: number[],
+  startedYears: Set<number>,
+): number {
+  const openYears = reportYears.filter((year) => startedYears.has(year));
+  if (openYears.length > 0) return Math.max(...openYears);
```
guilbep: @BenBotros  superweird that we have reportYears and StartedYears that define openYears?

the user should not access this data ?

I don't understand.

default year should be something like


function pickLatestYear(openYears: number[]): number {
  if (openYears.length === 0) {
    throw new Error('Invariant violation: expected at least one open report year');
  }

  return Math.max(...openYears);
}

we don't care about reportYears.. imho

guilbep (changes requested):

frontend/src/router/index.ts
```diff
@@ -51,17 +52,20 @@ export default route(function (/* { store, ssrContext } */) {
...
+ It renders nothing — its guard resolves a default unit/year and forwards
+ to the unified home page (/en/:unit/:year/home), or to /unauthorized when
+ the account has no units.
   ** LOGGED IN USERS **
-  / -> redirect to /:language(workspace-setup) with current locale
-  /en -> redirect to /en/workspace-setup
-  /en/login -> redirect to /en/workspace-setup
+  / -> redirect to /:language with current locale
+  /en -> resolve default workspace -> /en/:unit/:year/home
+  /en/login -> redirect to /en
```
guilbep: wrong comment, please fix

frontend/src/router/index.ts
```diff
@@ -8,6 +8,7 @@ import {
 
 import routes from './routes';
 import { authGuard } from './guards/authGuard';
+import workspaceGuard from './guards/validateUnitGuard';
```
guilbep: @BenBotros when you don 't need to, don't rename, if you want validUnitGuard to be renamed workspaceGuard, rename the file instead!

frontend/src/router/routes.ts
```diff
@@ -206,12 +213,25 @@ const routes: RouteRecordRaw[] = [
                 },
               },
               {
+                // The simulator home (Explore/Plan tabs) now lives inline on the
+                // unified home page; this route just redirects there, preserving
+                // the language/unit/year params.
                 path: 'simulation',
                 name: 'simulation',
-                component: () => import('pages/app/SimulationsPage.vue'),
+                redirect: (to) => ({
```
guilbep: @BenBotros soo? that's defensive programing? in case someone has a /simulation url somewhere? 
delete! don't keep dead-code, or fallback or legacy url. to early in the project for that

frontend/src/router/routes.ts
```diff
@@ -151,22 +167,13 @@ const routes: RouteRecordRaw[] = [
...
-              note: 'Workspace configuration - Year and lab selection',
-              breadcrumb: false,
-            },
-          },
           {
             path: `:unit(${UNIT_PATTERN})/:year(${YEAR_PATTERN})`,
             name: WORKSPACE_ROUTE_NAME,
-            beforeEnter: validateUnitGuard,
-            component: () => import('pages/app/WorkspacePage.vue'),
+            // Pass-through layout: the workspace is loaded by the global
```
guilbep, line 168: @BenBotros ok so we don't have the component and the guard anymore?

frontend/src/router/routes.ts
```diff
@@ -151,22 +167,13 @@ const routes: RouteRecordRaw[] = [
...
+            // `workspaceGuard`, so this parent only needs to host the child
+            // <router-view>.
+            component: { render: () => h(RouterView) },
```
guilbep: use real .vue page instead of render

frontend/src/router/routeNames.ts
```diff
@@ -14,11 +14,14 @@ export const LOGIN_ROUTE_NAME = 'login';
 export const LOGIN_TEST_ROUTE_NAME = 'login-test';
 export const LOGIN_ROUTES = [LOGIN_ROUTE_NAME, LOGIN_TEST_ROUTE_NAME];
 export const HOME_ROUTE_NAME = 'home';
-export const WORKSPACE_SETUP_ROUTE_NAME = 'workspace-setup';
 export const WORKSPACE_ROUTE_NAME = 'workspace-dashboard';
 export const UNAUTHORIZED_ROUTE_NAME = 'unauthorized';
 export const NOT_FOUND_ROUTE_NAME = 'not-found';
-export const DEFAULT_ROUTE_NAME = WORKSPACE_SETUP_ROUTE_NAME;
+export const AUTH_COMPLETE_ROUTE_NAME = 'auth-complete';
```
guilbep: auth-complete don't exist anymore, please remove

frontend/src/components/organisms/workspace-selector/WorkspaceSelectorBar.vue
```diff
@@ -0,0 +1,193 @@
...
+const selectedUnitId = computed({
+  get: () => selectedUnit.value?.id ?? null,
+  set: (id: number | null) => {
+    if (id != null) void handleUnitChange(id);
+  },
+});
+
+/**
+ * Years that are globally open (`is_started`) AND have a carbon report for the
+ * selected unit. Mirrors the intersection used by the old workspace-setup page.
```
guilbep: @BenBotros comment is wrong, you only use carbonReports below.
not sure why you do a Set though ?

imho return [...workspaceStore.carbonReports.map(x => x.year)] should be enough

frontend/src/components/organisms/workspace-selector/WorkspaceSelectorBar.vue
```diff
@@ -0,0 +1,193 @@
...
+  },
+});
+
+/**
+ * Pick the most recent open year for the freshly-loaded reports of a unit.
+ * Falls back to the latest report year, then to last calendar year.
+ */
+function defaultYear(): number {
+  const started = yearConfigStore.startedYears;
+  const reportYears = workspaceStore.carbonReports.map((report) => report.year);
```
guilbep: looks like duplicate of pickDefaultYear
same logic, should be WAY simpler than that

frontend/src/components/organisms/workspace-selector/WorkspaceSelectorBar.vue
```diff
@@ -0,0 +1,193 @@
...
+  const openYears = reportYears.filter((year) => started.has(year));
+  if (openYears.length > 0) return Math.max(...openYears);
+  if (reportYears.length > 0) return Math.max(...reportYears);
+  return new Date().getFullYear() - 1;
+}
+
+async function handleUnitChange(unitId: number) {
+  if (unitId === selectedUnit.value?.id) return;
+  const unit = workspaceStore.units.find((u) => u.id === unitId);
+  if (!unit) return;
```
guilbep, line 69: not sure why ?? we guard against bad unitId ? it's the role of the backend
frontend Should be DUMB

frontend/src/components/organisms/workspace-selector/WorkspaceSelectorBar.vue
```diff
@@ -0,0 +1,193 @@
...
+  await workspaceStore.fetchCarbonReportsForUnit(unitId);
+  const year = defaultYear();
+  await router.push({
+    name: HOME_ROUTE_NAME,
+    params: { ...route.params, unit: unitSlug(unit), year: String(year) },
+  });
+}
+
+async function handleYearChange(year: number) {
+  if (year === workspaceStore.selectedYear) return;
```
guilbep: only used in setter it seams.. plus we already guard in there? not sure why it's its specific function

my guess is that we should not guard against weird behavior

It's weird that we can set set: (year: number | null) => {
    if (year != null) void handleYearChange(year);
  },

year to null ?


handleUnitChange and HandleYearChange, is basically the same function and should BE the same function ? why encode unit each time?

either we change unit or year

function genericRouteUpdate(routeParams) {
await router.push({
    name: HOME_ROUTE_NAME,
    params: { ...route.params, ...routeParams},
  });
}

genericRouteUpdate({unit: unitSlug(unit)});
genericRouteUpdate({year: String(year) });

or something like this

frontend/src/components/organisms/workspace-selector/WorkspaceSelectorBar.vue
```diff
@@ -0,0 +1,193 @@
...
+const affiliationLeaf = computed(() => affiliationSegments.value.at(-1) ?? '');
+
+async function loadReports(unitId: number | undefined) {
+  if (unitId == null) return;
+  await workspaceStore.fetchCarbonReportsForUnit(unitId);
+}
+
+onMounted(async () => {
+  await Promise.all([
+    workspaceStore.units.length === 0
```
guilbep: why fail silently ?

frontend/src/components/organisms/workspace-selector/WorkspaceSelectorBar.vue
```diff
@@ -0,0 +1,193 @@
...
+    name: HOME_ROUTE_NAME,
+    params: { ...route.params, unit: unitSlug(unit), year: String(year) },
+  });
+}
+
+const affiliationSegments = computed(
+  () => selectedUnit.value?.affiliations ?? [],
+);
+// Parent path (greyed), with trailing separator when there's a leaf after it.
+const affiliationPrefix = computed(() => {
```
guilbep: PURE CSS possible ?

frontend/src/pages/app/HomePage.vue
```diff
@@ -1,27 +1,38 @@
...
 const yearConfigStore = useYearConfigStore();
+const { t, te, locale, getLocaleMessage } = useI18n();
+
+/** A translation key resolves to displayable content (exists and non-empty). */
+const hasText = (key: string) => te(key) && t(key).trim().length > 0;
 
-const hasModulePermission = authStore.hasUserModulePermission;
+/** Keeps the ECharts-heavy bundle out of the initial home route chunk. */
+const ChartSkeleton = () =>
+  h(QSkeleton, { type: 'rect', height: '360px', class: 'full-width' });
```
guilbep: @BenBotros no h() only pure template please as much as possible

frontend/src/pages/app/HomePage.vue
```diff
@@ -46,289 +74,512 @@ const firstEditableModule = computed(() =>
...
+  return `mailto:${email}?subject=${encodeURIComponent(
+    subject,
+  )}&body=${encodeURIComponent(body)}`;
+});
 
-type StatusBadgeVariant = 'validated' | 'progress';
+const calculatorUpdates = computed(() => {
+  // Discover update indices from the i18n keys (calculator_update_<n>_*) instead
+  // of hard-coding a count, so adding entries only requires editing the locale file.
+  const indices = [
```
guilbep, line 106: 👍

frontend/src/router/guards/redirectToDefaultRoute.ts
```diff
@@ -0,0 +1,128 @@
+import type { RouteLocationNormalized } from 'vue-router';
+import { useWorkspaceStore, unitSlug } from 'src/stores/workspace';
+import { useYearConfigStore } from 'src/stores/yearConfig';
+import { useAuthStore, PermissionAction } from 'src/stores/auth';
+import { resolveNoUnitRoute } from 'src/utils/unauthorized';
+import { currentLanguage } from 'src/utils/language';
+import { HOME_ROUTE_NAME } from '../routeNames';
+
+/**
+ * `workspaceGuard` redirects here with `?unit&year` (both null) when a persisted
```
guilbep: @BenBotros does workspaceGuard even exist anymore? It's all very confusing

frontend/src/router/guards/redirectToDefaultRoute.ts
```diff
@@ -0,0 +1,128 @@
...
+) {
+  const workspaceStore = useWorkspaceStore();
+
+  if (isForcedReset(to)) {
+    workspaceStore.reset();
+  }
+
+  // Resolve the user's units up front. A no-unit account must always be routed
+  // by unit membership, never by a persisted selection: `selectedParams` is
+  // persisted to localStorage, so a stale value from a prior session would
```
guilbep: too complicated! we need something dumber

guilbep (changes requested):

backend/app/api/v1/auth.py
```diff
@@ -551,18 +553,26 @@ async def oauth_callback(
...
     identity (institutional_id, provider) from JWT. Uses cached DB
     roles — does not sync from the role provider synchronously.
+
+    Beyond the user, the response bundles the units the caller can access and
+    the globally-configured years, so the frontend hydrates its whole auth/
+    workspace context in a single request instead of three.
     """
+    # Local import avoids any import-order coupling between the auth and
+    # year-configuration route modules (both are pulled into the API router).
+    from app.api.v1.year_configuration import list_configured_years
```
guilbep: @BenBotros  toujours faire des imports top of file, no inline

backend/app/api/v1/auth.py
```diff
@@ -551,18 +553,26 @@ async def oauth_callback(
 # ---------------------------------------------------------------------------
 
 
-@session_router.get("", response_model=UserRead, response_model_exclude_none=True)
+@session_router.get("", response_model=SessionRead, response_model_exclude_none=True)
```
guilbep, line 557: @ymarcon qu'en penses-tu ?

backend/app/api/v1/carbon_report_module_stats.py
```diff
@@ -83,6 +68,31 @@ async def get_validated_totals(
     )
 
 
+@router.get("/{carbon_report_id}/validated-totals")
```
guilbep: @BenBotros dead-code?

backend/app/api/v1/year_configuration.py
```diff
@@ -480,6 +480,28 @@ async def create_audit_entry(
...
+async def list_configured_years(
+    db: AsyncSession, current_user: User
+) -> list[YearConfigurationListItem]:
+    """List year-configuration rows visible to ``current_user``.
+
+    Extracted from the ``GET /`` route so the enriched ``GET /session``
+    (bootstrap) can embed the workspace year selector's rows without a second
+    HTTP round trip. See ``list_year_configurations`` for the filtering rules.
+    """
+    is_admin = await is_permitted(current_user, "backoffice.configuration", "view")
```
guilbep: @BenBotros check if admin + principal, do you see  all configuration year?

backend/app/api/v1/workspace_home.py
```diff
@@ -0,0 +1,99 @@
...
+    Gets (or creates) the carbon report for ``unit_id``/``year``, then bundles
+    the year configuration and the emission breakdown (with the validated-only
+    total merged in; the breakdown also carries the per-module states).
+    """
+    unit = await db.get(Unit, unit_id)
+    require_unit_access(current_user, unit)
+
+    report_service = CarbonReportService(db)
+    report = await report_service.get_by_unit_and_year(unit_id, year)
+    if report is None:
```
guilbep, line 75: @BenBotros on arrête de créer les reports

backend/app/api/v1/workspace_home.py
```diff
@@ -0,0 +1,99 @@
...
+class WorkspaceHomeResponse(BaseModel):
+    """Minimal aggregate payload the frontend fans out across its stores."""
+
+    carbon_report_id: int
+    year_config: Optional[YearConfigurationResponse] = None
+    emission_breakdown: dict
+
+
+@router.get("/{unit_id}/{year}/home", response_model=WorkspaceHomeResponse)
+async def get_workspace_home(
```
guilbep, line 58: @BenBotros get is READ ONLY

backend/app/api/v1/carbon_report_module_stats.py
guilbep: 🧏‍♂️

guilbep (changes requested):

frontend/src/router/guards/redirectToDefaultRoute.ts
```diff
@@ -0,0 +1,65 @@
...
+ * the user's first unit + most recent open year — and forward to the home page,
+ * or to {@link resolveNoUnitRoute} (back-office for admins, /unauthorized
+ * otherwise) when the account has no units.
+ */
+export default async function redirectToDefaultRoute(
+  to: RouteLocationNormalized,
+) {
+  const workspaceStore = useWorkspaceStore();
+
+  await workspaceStore.getUnits();
```
guilbep: already done in /session 
not useful anymore! SIMPLIFY

guilbep (changes requested):

guilbep:
@BenBotros Claude il dit:


Findings (most severe first)

1. frontend/src/router/guards/workspaceGuard.ts:91 — loadWorkspaceFromRoute only redirects to the landing resolver when the unit itself is invalid, not when the new aggregate call fails.
fetchWorkspaceHome (workspace.ts) catches all errors and returns null, so carbonReportId stays null while response (unit validity) is still true. The guard exit check if (!response && !carbonReportId) evaluates false, so navigation proceeds into Home/Module/Results with the timeline, year-config, and module stores never hydrated — a transient 500/network blip on this one aggregate call silently blanks three stores instead of blocking navigation.

2. frontend/src/pages/ErrorUnauthorized.vue:46 + frontend/src/router/guards/redirectToDefaultRoute.ts:55 — users redirected here with reason=no-open-year (no globally-open year) hit a dead end.
isNoUnit only special-cases reason === 'no-unit'; for no-open-year the page shows a "Home" button instead of the Logout escape hatch the no-unit case gets. Clicking "Home" re-enters redirectToDefaultRoute, which again finds startedYears.size === 0 and returns resolveNoOpenYearRoute() — bouncing right back to the same page with no way out.

3. frontend/src/components/organisms/workspace-selector/WorkspaceSelectorBar.vue:69 — handleUnitChange calls pickDefaultYear(yearConfigStore.startedYears) without checking for an empty set, violating that function's own documented precondition ("Callers must pass a non-empty set"). If all globally-open years close while a user's Home tab stays open, switching units computes Math.max() over an empty set → -Infinity, producing a route push to /{unit}/-Infinity/home.

4. frontend/src/stores/modules.ts:1004 + frontend/src/pages/app/HomePage.vue:41-48 — editing any module regresses the Home headline total to a dash.
total_tonnes_validated_co2eq is documented as "absent when the breakdown comes from the plain /emission-breakdown route" (only the workspace-home aggregate merges it in). refreshEmissionBreakdownIfNeeded, called after every module save, invalidates the cache and refetches via that plain route — overwriting the richer aggregate payload. HomePage.vue's totalTonnesCo2 reads total_tonnes_validated_co2eq directly, so it becomes undefined and renders "-" the next time the user returns to Home, even though hasValidatedData (keyed off validated_categories) still correctly shows the chart.

5. frontend/src/router/guards/redirectToDefaultRoute.ts:15,41 — unconditionally calls workspaceStore.getUnits() and yearConfigStore.fetchConfiguredYears() even though authGuard (which runs first in the guard chain) already awaits auth.bootstrap(), which hydrates both via the new single session payload. Every hit of the parameterless landing route (post-login, refresh, back-button) fires two redundant network calls that duplicate data the bootstrap just delivered a moment earlier.

6. frontend/src/components/charts/results/ModuleCarbonFootprintChart.vue:136-139 — CATEGORY_TO_SUBMODULE hand-inverts SUBMODULE_TO_CATEGORY     (already defined in useModuleIconColors.ts:20-23), linke SUBMODULE_TO_CATEGORY"). Adding or renaming a submodulecategory now requires editing two maps in opposite directions by hand; missing one silently mis-colors or mis-links an icon-axis button with no compile-time error. Three independent review passes flag
                                                                                                                                                7. frontend/src/stores/workspace.ts:308 — reset() (clearnitResults/selectedParams) is now dead code — its onlycaller was the deleted redirectToWorkspaceIfSelectedGuard flow, and nothing in the new guard chain calls it. Confirmed via repo-wide grep: zero remaining call sites.
                                                                                                                                              8. frontend/src/i18n/home.ts (~25 keys, e.g. workspace_r, workspace_workspace_title, workspace_simulator_*) —copy-pasted verbatim from the deleted workspace_setup.ts under the new prefix, but referenced nowhere now that WorkspaceSetupPage.vue is gone (confirmed via repo-wide grep). Dead translation weight slated for strings nobody renders.

guilbep (approved):
il m'a eu à l'usure

PR #1714, opened 2026-07-05
fix(1575): emit a single heating leaf (electric OR thermal) per room

### LINKED ISSUE #1575: [BUG] (Building) ROOMS: calculation for centralized heating seems wrong
While testing the upload of different factors for buildings rooms, we realized that the visual on centralized heating vs electric heating always shows the same quantity for both, which should not be the case. 

Can we double check how you implemented this? 

The behaviour I expect with the factors I uploaded now is to have 0 centralized heating and 100% electric heating in this graph, which is not the case: 

Image

@Anna-Kounina-Masse fyi
@charlottegiseleweil

guilbep:
## Why

PR #1661 (issue #1575) split building-room heating into electric vs thermal leaves, but a room's `kwh/m²` was emitted against **both** leaves and the mismatched one was zeroed via an overloaded `conversion_factor = 0`. That approach:

- emitted a spurious `0.0` row for the non-applicable heating leaf;
- overloaded `conversion_factor` as a kill-switch (and coerced a legitimate `0.0`, e.g. a carbon-free network, to `1.0` via a falsy `or`);
- silently dropped **all** heating at ZZ level (missing/invalid `room_type`).

## What

Select the single correct heating leaf up front from the matched factor's `energy_type`, so a room emits **electric OR thermal — never both**.

- `energy_type` moved back to the factor's **classification** fields; alias-normalization dropped (values are already `electric`/`thermal`).
- New `DataEntryEmissionService._get_building_energy_type` resolves the matched factor's `energy_type` (cache-aware for bulk recalc, single DB `get` otherwise) and **fails loudly** on a missing/corrupt factor rather than silently dropping heating.
- `resolve_emission_types` takes an explicit `building_energy_type` and dispatches building rooms through `_resolve_building_rooms`, which appends only the matching `heating_electric` / `heating_thermal` leaf (`None` → no heating). `data` is never mutated; the energy type is passed explicitly, not smuggled through the payload.
- `EmissionType.buildings__rooms__heating_elec` → `heating_electric`; frontend chart constants updated to match.
- `_compute_kwh_emission` simplified: `conversion_factor` applies to heating only, defaults to `1.0`, and an explicit `is None` check preserves a legitimate `0.0`.

## Result

A room with only an electric factor now emits **0 thermal / 100% electric** (and vice-versa); the rollup no longer double-counts (65.0, not 95.0).

## Tests

- Unit: `test_buildings_schemas.py` (formula + `resolve_computations` selection), new `test_building_rooms_resolver.py`, `test_data_entry_emission_service.py` energy-type resolution.
- Integration: `test_buildings_csv_pg.py` — electric factor emits only the electric leaf, thermal factor only the thermal leaf, no double-count; `test_strategy_b_rematch_pg.py` updated.

Closes #1575

### DIFF
--- a/backend/app/services/data_entry_emission_service.py
+++ b/backend/app/services/data_entry_emission_service.py
@@ -23,6 +23,7 @@ from app.repositories.data_entry_emission_repo import (
 from app.schemas.data_entry import BaseModuleHandler, DataEntryResponse
 from app.services.factor_service import FactorService
 from app.utils.data_entry_emission_type_map import (
+    BUILDING_ENERGY_TYPES,
     DATA_ENTRY_TYPE_TO_ROLLUP_EMISSION,
     resolve_emission_types,
 )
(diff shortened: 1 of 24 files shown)


backend/app/services/data_entry_emission_service.py
```diff
@@ -226,6 +227,40 @@ async def _get_percentage_override_kg(
 
         return prev_kg * (percentage / 100.0)
 
+    async def _get_building_energy_type(
```
ymarcon, line 230: This is building specific, then should be in the building's module, not in this generic service.

guilbep:
Superseded by #1720, which tracks fix/1661-refactor-generic (includes the generic factor-hook refactor). fix/1661-refactor is kept as a pre-refactor backup.

PR #1719, opened 2026-07-06
refactor(1661): remove primary_factor_id from DataEntry.data, replace-semantics factor reupload

guilbep:
Closes the `MultipleResultsFound` 500 on factor lookups at its root by removing the `primary_factor_id` denormalization from `DataEntry.data` entirely.

**Stacked on #1714 (`fix/1661-refactor`)** — merge that first.

Spec: `docs/src/implementation-plans/1661-remove-primary-factor-id.md` · Plan/execution log: `docs/src/implementation-plans/1661-remove-primary-factor-id-plan.md`

## Phase 1 — factors are derived state

- New `app/services/factor_resolver.py`: on-demand, memoized resolution from the entry's classification fields (one bulk SELECT per `(data_entry_type, year)` per request/recalc-slice — accepted robustness-over-performance trade-off). Owns the falsy-kind short-circuit before any load.
- Every writer/reader of `data["primary_factor_id"]` is gone: emission compute injects the resolved id into `ctx`; the recalc workflow lost its whole rematch machinery (file 500→267 lines) and never mutates `entry.data`; API create no longer resolves at all; CSV ingest stops stamping; list enrichment resolves per row (ambiguity-tolerant: logs + empty factor columns instead of a page-wide 500).
- `DataEntryEmission.primary_factor_id` (real FK) is unchanged — it records what was actually used.

## Phase 2 — replace-semantics factor reupload

- `FactorRepository.delete_stale_for_year(year, *, det_ids, threshold_job_id)`: the factor CSV provider deletes superseded rows right after a successful upsert, same transaction, before the recalc fan-out; emission rows cascade and are rebuilt by the chained recalc.
- **Destructive-semantics guards:** sweep runs only on full-SUCCESS uploads (a partial upload never deletes rows it failed to re-ingest), and only for the dets the job actually upserted (a partial module CSV cannot wipe sibling dets). Both unit-pinned.
- `GET /v1/factors/stale`, `list_stale_for_year`, `_latest_factor_job_per_det` and the per-det SQL CASE threshold are deleted; openapi regenerated (snapshot refreshed from this branch's schema).
- Regression test reproduces the originating bug (building_rooms 2-key → 3-key classification reshape → `MultipleResultsFound`) and proves the sweep kills it.

## Bugs found & fixed along the way

- **Silent emission wipe:** PATCH blanking/nulling `purchase_institutional_code` briefly turned from a loud 400 into a 200 that deleted the entry's emissions — now rejected at validation (both flat and data-nested payload shapes), regression-tested.
- Latent test bug: a CSV-provider test's factors_map key missed the year segment, so its match path never ran.

## Behavior changes to note

- `prepare_create`'s `data_entry.id is None` guard moved to the top: unpersisted entries with corrupt classification now return `[]` (logged) instead of raising; all production callers flush first.
- Report exports no longer scrub `primary_factor_id` from entry JSON — pre-branch rows on non-reseeded DBs expose the stale key until the next reseed (v0.x drops the DB).
- `PurchaseHandlerCreate` accepts whitespace-only codes while update rejects them; a legacy entry persisted with a null code would 400 on every PATCH until reseeded (moot post-reseed).

## Verification

- Unit: 1756 passed; ruff + mypy clean.
- Integration: 348 passed; the residual 2 failures + 5 errors reproduce on the pre-branch baseline (local ES-container/environment, verified against 9f2d6bb8).
- Frontend: `openapi.d.ts` regenerated, `make -C frontend type-check` green.
- Reviews: per-task review gates + two whole-branch reviews (Phase 1, Phase 2); all Critical/Important findings fixed and re-verified.

## Follow-ups (not this PR)

- `_detach` resolver-loaded factors in the list-enrichment fallback (defense-in-depth symmetry).
- Split `recalculate_for_data_entry_type` (~230 lines) and `prepare_create` (>40-line rule, pre-existing).
- Single-entry PATCH does two bulk factor loads (handler-service resolver + prepare_create resolver) — acceptable, could thread one resolver later.


guilbep:

PR #1722, opened 2026-07-06
fix(1661): v1.0 data migration — strip legacy entry-json keys

guilbep:
**Stacks on #1719** (merge into it or after it — one commit).

At v1.0 the DB persists across deploys, so two #1719 decisions that leaned on "moot after the next reseed" need a real data migration (`954eac6c95da`, data-only, no-op downgrade):

- strips the dead `primary_factor_id` key from `data_entries.data` (nothing reads it since #1719; report exports would expose it)
- strips explicit-null `purchase_institutional_code` keys — under #1719's validator those rows would 400 on **every** PATCH, i.e. become permanently un-editable

Also updates the 1661 spec/plan docs accordingly and records the post-deploy operator step: **re-upload the current factor CSVs per module/year** so the replace-semantics sweep deletes stale factor generations that accumulated in prod before #1719 (the sweep only fires on new uploads; without this, pre-existing 2-key/3-key duplicates keep 500-ing `/values`).

Verified: `test_alembic_migrations.py` green; ruff clean.

Note: generated with plain `alembic revision` (no `--autogenerate`) — a data-only migration has no schema diff, and autogenerate would only produce false positives to prune.

### DIFF
--- a/backend/alembic/versions/2026_07_06_1124-954eac6c95da_strip_legacy_primary_factor_id_and_null_.py
+++ b/backend/alembic/versions/2026_07_06_1124-954eac6c95da_strip_legacy_primary_factor_id_and_null_.py
@@ -1,0 +1,62 @@
+# codeql[py/unused-global-variable]
+"""strip legacy primary_factor_id and null purchase codes from entry json
+
+Revision ID: 954eac6c95da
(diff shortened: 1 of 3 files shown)


guilbep:

PR #1724, opened 2026-07-06
fix(1721): sort/search coverage — 6 module map gaps + matrix regression test

guilbep:
Stacks on #1721. Fixes found while testing:

- **buildings**: surface sort now reads the joined `building_rooms` row (was all-NULL from entry data)
- **purchase**: `currency` added to sort_map (was 'Cannot sort by unknown field')
- **research facilities** (common + animals): full sort_map/filter_map (was id+kg only / empty)
- **energy_combustion, building_embodied_energy, external_clouds, external_ai, process_emissions**: Factor-only sort/filter expressions now coalesce with entry data — search finds what the row displays even when no factor matched
- **count query**: joins Factor exactly like the page query per branch (was an implicit cross join for Factor-referencing filters → 0 or inflated totals; found by the matrix on embodied)

New `test_submodule_sort_search_matrix_pg.py`: drives every registered handler with synthesized entries and asserts **every sort key (asc+desc), every filter key, and pagination** — new map keys are covered the moment they're added. Plus a targeted buildings-surface regression test.

Verified: unit 1747 + ruff + mypy green; matrix + sort integration files green.

### DIFF
--- a/backend/app/modules/buildings/schemas.py
+++ b/backend/app/modules/buildings/schemas.py
@@ -6,6 +6,7 @@ from pydantic import (
     ValidationInfo,
     field_validator,
 )
+from sqlalchemy import func
 
 from app.core.logging import get_logger
 from app.models.data_entry import DataEntry, DataEntryTypeEnum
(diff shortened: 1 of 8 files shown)


guilbep:

PR #1739, opened 2026-07-07
perf(1401): openshift stage resource quota investigation

guilbep:
Part of #1401.

Implementation plan: `docs/src/implementation-plans/1401-openshift-resource-optimization.md`

_Branch scaffold only — this issue needs reporter/DevOps clarification before a fix can be scoped, see plan._


guilbep:
not gonna implement

Issue #1433, opened 2026-06-08
[Bug]: (Back-office) configuration - Headcount Module


CarolineChereau:
### What went wrong?

1- you  add data in the configurator : Headcount, Equipment, Processs Emission, Professional travel
2-you desactivate "Headcount" and the you can show that all data in others modules are deleted.

It's a logical behavior for Professional travel (linked to Headcount) but not with the others module. I think, an message when you want to deactivate headcount module is to inform what does it mean for others module, or not authorize to deactivate the Headcount Module 

### How can we reproduce this?

1.Go to Configurator 
2- Add data in several Module (Headcount, travel, equipment, process emissions, purchases)
3- Deactivate Headcount Module
4-Show data in the module where you added data

### Screenshot

@martina-gallato for information

### Which browser are you using?

Chrome

### What device are you using?

Desktop/Laptop

### Before submitting

- [x] I've checked that this issue hasn't already been reported

guilbep:
@CarolineChereau I don't understand exactly the problem. could you give us more context? maybe screenshot ? and a clear behavior please? I'm moving this issue back to definition

CarolineChereau:
@martina-gallato est-ce que ce problème est toujours présent ?

charlottegiseleweil:
@martina-gallato ?

martina-gallato:
all good!

CarolineChereau:
@charlottegiseleweil I close the issue

PR #1742, opened 2026-07-07
fix(1433): headcount deactivation hides unrelated module data

guilbep:
Part of #1433.

Implementation plan: `docs/src/implementation-plans/1433-headcount-deactivation-deletes-other-modules-data.md`

Root-cause mechanism wasn't pinned down by static review alone — plan leads with a reproduction test and a UX stopgap (confirmation dialog), see plan for details.

_Branch scaffold only — plan not yet implemented._


guilbep:
seems like it's all good

PR #1744, opened 2026-07-07
refactor(1491): factor identity-key hardening, viewer, bulk delete

guilbep:
Part of #1491.

## Re-scope

The duplicate-factor bug that motivated #1491 is structurally fixed by the 310B upsert + stale sweep, and plan 1661 removed the original reason id-stability was load-bearing. Per the maintainer decisions (2026-07-10), the upsert **stays** and gets hardened; this PR delivers the re-scoped asks 2–4 plus the ask-1 residual documentation. Full plan: `docs/src/implementation-plans/1491-factor-upsert-strategy.md` (status: delivered).

## What's in here

**Ask 1 residual — documentation**
- Corrected the stale rationale comments in `factor_repo.upsert_factors` and `base_factor_csv_provider` (they cited the removed persisted-FK; the real rationale is partial-upload safety + blast radius).
- New operator-facing page `docs/src/backend/12-FACTOR-LIFECYCLE.md`: upsert semantics, SUCCESS-only stale sweep, chained recalc, what re-importing a corrected CSV does.
- **Sweep policy decision (option 1, documented only)**: the factor CSV is the source of truth for its `(det, year)` scope; non-CSV factor writes (computed providers) are transient by design. No sweep-predicate change.

**Ask 2 — null-identity-field guard**
- `_process_row` now rejects rows whose DTO-required classification fields are null with a per-row error naming the field(s), before emission-type resolution and before the null can enter the upsert identity key. Requiredness derives from the handler DTO's required-column set, so legitimately nullable classification fields still pass.
- Regression tests: row error naming the field, WARNING result, no factor written, no stale sweep.

**Ask 3 — backoffice factor viewer**
- `GET /api/v1/backoffice/factors?data_entry_type_id=&year=&page=&page_size=` — paginated, rows serialized through the type's factor handler plus `year` and `last_seen_job_id` (spot rows the latest upload did not assert). Gated `backoffice.configuration:view`.
- Frontend: read-only server-paginated `q-table` section on the Data Management page (reuses the page's year selector, local data-entry-type select); `api/factors.ts` + factors store extended; en/fr i18n keys.

**Ask 4 — bulk delete endpoints**
- `DELETE /api/v1/backoffice/factors?data_entry_type_id=&year=` (list-ids + bulk delete composition).
- `DELETE /api/v1/backoffice/data-entries?data_entry_type_id=&source=&year=[&carbon_report_module_id=]` wired to the existing `DataEntryService.bulk_delete_by_source` (module-scoped, audited) / `bulk_delete_by_source_year`. A module-scoped delete validates the module's year matches the param.
- Both gated `backoffice.configuration:edit`; both return the deleted count plus the dispatched recalc job/pipeline ids.

### Recalc vs recompute-stats after a bulk delete (decision)

Each delete dispatches a **root `emission_recalc` job** (same shape as `POST /sync/recalculate-emissions/{module}/{det}`), not the admin `recompute-stats` trigger. Reasoning:

- Deleting factors CASCADE-deletes the emission rows that reference them, but emission rows with a NULL `primary_factor_id` survive with stale values, and surviving data entries need their emissions **rebuilt**, not just re-counted. Only the recalc path does that (set-based replace per entry), and its handler chains the trailing `aggregation` that rewrites `carbon_report_module.stats` — so the stat buckets can't go stale.
- `recompute-stats` only re-aggregates existing emission rows and is designed for unchanged data (skips the module-status bump, skips scopes without a current FACTORS job) — it would preserve stale emission rows in the stats.
- Module-scoped data-entry deletes pin `carbon_report_module_ids` on the recalc config, mirroring the unit-specific ingest, so a small delete doesn't recompute the whole `(det, year)` slice.

## Verification

- Backend: `tests/unit/services/data_ingestion` + `tests/unit/v1` (472 passed) and `tests/unit/{repositories,tasks,workflows}` (389 passed); `ruff check` / `ruff format --check` / `mypy app/` clean.
- Frontend: `vue-tsc --noEmit` clean; eslint + prettier clean on touched files.
- Docs: prettier clean.


QuentinEschmann:
@enac-ai-reviewer explain

guilbep:
Closing — the feature half is declined, but **one piece here is a real bug fix and should be salvaged.**

#1491 is closed and the view/delete mechanism is not being built, which retires **ask 3** (backoffice factor viewer) and **ask 4** (bulk-delete endpoints). That is most of this PR's 1423 lines and all of its frontend surface. The branch is also `CONFLICTING` against `dev` after two months.

### Worth keeping: the null-identity-field guard (ask 2)

```
# a null in a *required* classification field must be a per-row error, never a
# silently-null component of the create DTO
```

Verified against `dev` today: `required_columns` still exists on the handlers, and **the guard is not there.** So the hole is live.

It matters for the same reason everything else this quarter did: a null in a required classification field becomes a null component of the **upsert identity key**. Rows then collide or duplicate on a key nobody intended, silently, and the factor set the emissions are computed from is quietly wrong. That is the #2546 shape, one layer further upstream.

It ships with three regression tests, including one pinning that such a row yields a WARNING result and **no stale sweep** — which is the part that would otherwise delete good factors on the strength of a bad upload.

### Also cheap to keep

- The corrected rationale comments in `factor_repo.upsert_factors` / `base_factor_csv_provider` — they cited a persisted FK that no longer exists, so they actively mislead.
- `docs/src/backend/12-FACTOR-LIFECYCLE.md` — operator-facing: upsert semantics, SUCCESS-only stale sweep, chained recalc, what re-importing a corrected CSV does. Useful independently of whether anyone builds a viewer.

### Recommendation

Extract ask 2 plus those two documentation pieces into a small PR against a fresh bug issue — roughly 60 lines of app code and three tests, instead of 1423 lines and a UI. The rest of this branch stays closed with #1491.

Happy to do that extraction; say the word.

PR #1811, opened 2026-07-15
fix: updated recompute missing factors from animal facilities

### LINKED ISSUE #1597: [BUG] (Research facilities) animal factors do not behave as expected
By trying out different values in research_facilities_animal_factors we realized the behaviour is not what we expected. 

Expected behaviour: 

- If one of the columns in `kg_co2eq_sum_processemissions`,  `kg_co2eq_sum_building_energycombustions`, `kg_co2eq_sum_building_rooms`, `kg_co2eq_sum_purchases_common` , `kg_co2eq_sum_purchases_additional`, `kg_co2eq_sum_equipments` is filled, this values needs to be used for the corresponding type of animal facility (mice vs fish) and the corresponding module/submodule. This implies that if these values are given, since they already belong to one of the two types of animal facility, the columns `module_submodule_share` are not used in the calculation. 

- If the value for `kg_co2eq_sum_module/submodule` is not given, the calculator needs to compute this value for the animal facility (in a similar way to the common_research_facility compure missing factors button), and then take only the share that is assigned to the type in question, which is given in the column `module_submodule_share`. 

I think right now this is not what was implement, in particular it seems that now what happens is the share is always considered, and multiplied by the value given in the kg_co2eq. This also means that if this value is left empty, the result is 0. 

Can you fix this? 

@maidinh-metier fyi
@charlottegiseleweil

ymarcon:
## What does this change?

* Apply emission share when recalculating animal facilities factors
* Added compute missing factors button to animal facilities submodule config

## Type of change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📝 Documentation update
- [ ] 🎨 Design/UI improvement
- [ ] 🔧 Configuration change
- [ ] 🧹 Code cleanup

## Code quality checklist 

- [ ] No hardcoded values or secrets
- [ ] Documentation updated for new features
- [ ] Commit messages follow convention
- [ ] No console.log or debug statements

## Testing checklist

## Related issues

- Closes #1597 

---

### DIFF
--- a/backend/app/modules/research_facilities/animals_schemas.py
+++ b/backend/app/modules/research_facilities/animals_schemas.py
@@ -162,7 +162,7 @@ class ResearchFacilitiesAnimalModuleHandler(BaseModuleHandler):
                     # Source emissions are missing in the factor values
                     researchfacility_name = ctx.get("researchfacility_name", "Unknown")
                     researchfacility_type = ctx.get("researchfacility_type", "Unknown")
-                    logger.warning(
+                    raise ValueError(
                         f"Missing kg_co2eq_sum for source {source} in factor values "
                         f"for facility {researchfacility_name} "
                         f"({researchfacility_type})."
--- a/backend/app/modules/research_facilities/animals_schemas.py
+++ b/backend/app/modules/research_facilities/animals_schemas.py
@@ -130,7 +130,6 @@ class ResearchFacilitiesAnimalModuleHandler(BaseModuleHandler):
             """
             use = ctx.get("use")
             use_unit = ctx.get("use_unit")
-            kg_co2eq = ctx.get("kg_co2eq")
             if use is None or use_unit is None:
                 return None
             # Must be same unit for the factor and the data entry to apply the formula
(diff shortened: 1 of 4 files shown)


backend/app/modules/research_facilities/animals_schemas.py
```diff
@@ -153,21 +152,21 @@ def _research_facilities_formula(
...
+                if kg_co2eq_sum_source is not None:
                     # Add provided facility emissions modulo share of use
                     if kg_co2eq_sum is None:
                         kg_co2eq_sum = 0
                     kg_co2eq_sum += use_share * kg_co2eq_sum_source
+                else:
+                    # Source emissions are missing in the factor values
+                    researchfacility_name = ctx.get("researchfacility_name", "Unknown")
+                    researchfacility_type = ctx.get("researchfacility_type", "Unknown")
+                    logger.warning(
```
ymarcon: @guilbep Not sure what strategy is recommended, this is just a log but it could be a raised error.

guilbep: The problem is that kg_co2eq_sum is optional because it is supposed to be recalculated by the recompute_factors button. So, in theory, if it is not present, it means we forgot to run recompute_factors, which should result in a critical failure and raise an exception, such as a ValueError.

guilbep (changes requested):

